PR #1, opened 2017-05-09
Update package.json

UdoW:
Updates version of vscode language client to 3.1.0.


akosyakov:
@UdoW we use our own branch of `vscode-languageclient` with a vscode independent `BaseLanguageClient`. There is an issue for it in `vscode-languageserver-node`: https://github.com/Microsoft/vscode-languageserver-node/issues/88#issuecomment-299813935. But untill it is not resloved we cannot accept this PR.

UdoW:
@akosyakov , ok, I will follow your changes and do an upgrade later when everything fits.

PR #6, opened 2017-07-17
spread the command arguments

svenefftinge:
Signed-off-by: Sven Efftinge <sven.efftinge@typefox.io>


svenefftinge:
Command arguments currently get wrapped in another array

Issue #8, opened 2017-07-20
Example build fails with Node.js 8


franzbecker:
I cannot build the example on my machine with:
```
$ node --version
v8.2.0
$ npm --version
5.3.0
```

That seems to be a problem when using a newer Node.js / NPM version as you can see on Travis:
https://travis-ci.org/franzbecker/monaco-languageclient/jobs/255741802

The error occurs while compiling TypeScript:
```
> tsc

src/client.ts(60,57): error TS2345: Argument of type 'MessageConnection' is not assignable to parameter of type 'MessageConnection'.
  Types of property 'sendRequest' are incompatible.
    Type '{ <R, E, RO>(type: RequestType0<R, E, RO>, token?: CancellationToken | undefined): Thenable<R>; <...' is not assignable to type '{ <R, E, RO>(type: RequestType0<R, E, RO>, token?: CancellationToken | undefined): Thenable<R>; <...'. Two different types with this name exist, but they are unrelated.
      Types of parameters 'type' and 'type' are incompatible.
        Type 'RequestType0<any, any, any>' is not assignable to type 'RequestType0<any, any, any>'. Two different types with this name exist, but they are unrelated.
          Types have separate declarations of a private property '_'.
```

Phantomape:
Any idea how to solve it ?

franzbecker:
Haven't had any time to research it any further, but it might be a TypeScript issue, see this discussion: https://github.com/Microsoft/TypeScript/issues/16985 

The project should definitely use NPM 5 or Yarn in order to tie down the dependencies that are used for the build.

franzbecker:
Didn't have the time to research it any further, a quick Google searched revealed this issue: https://github.com/Microsoft/TypeScript/issues/16985 but not sure if it applies.

The project should definitely use NPM 5 or Yarn in order to tie down the dependencies that are used for the build.

Phantomape:
I changed "import { listen, MessageConnection } from 'vscode-ws-jsonrpc';" to "import { MessageConnection } from 'vscode-jsonrpc'; import { listen } from 'vscode-ws-jsonrpc';", and somehow got it working. Not sure that is the actual reason behind this error cause I also have some other modifications besides this change. As I dig deeper into the package vscode-ws-jsonrpc, I found out that MessageConnection in it is actually imported from vscode-jsonrpc, which confuses me; because as I see it, importing from vscode-jsonrpc or vscode-ws-jsonrpc doesn't make any difference.

[...]

PR #9, opened 2017-07-20
Integrate with Travis in order to build with node 7 and 8

franzbecker:
There should be a CI system for building this package that tests it with Node.js 8 and NPM 5 since it seems so fail (see issue https://github.com/TypeFox/monaco-languageclient/issues/8).

I like Travis so I created a .travis.yml (after merging this it would need to be enabled at https://travis-ci.org. This could also be used for an automated release process as a next step.
You can see how it looks like here: https://travis-ci.org/franzbecker/monaco-languageclient/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: node_js
+node_js:
+  - "6"
+  - "7"
+cache:
+  directories:
+    - node_modules
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "vscode-base-languageclient": "^0.0.1-alpha.2"
   },
   "scripts": {
-    "prepublish": "npm run clean && npm run compile",
+    "prepare": "npm run clean && npm run compile",
     "compile": "tsc",
     "watch": "tsc -w",
     "clean": "rimraf lib"


.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: node_js
+node_js:
+  - "7"
+  - "8"
+cache:
+  directories:
+    - node_modules
+script:
+  - cd example && npm install && npm run build
```
akosyakov: Could you remove `script` and replace `prepublish` npm script with `prepare`? travis by default runs `npm install` + `npm test` for `node_js` language.

.travis.yml
```diff
@@ -0,0 +1,9 @@
+language: node_js
+node_js:
+  - "7"
+  - "8"
```
akosyakov: The build for node 8 does not seem to work, could we remove it for now or see why it does not work?

akosyakov: It would be actually to build with node_js 6 as vscode-jsonrpc. And the bug with node 8 should be fixed first there.

.travis.yml
franzbecker: done

.travis.yml
franzbecker: done

akosyakov:
@franzbecker changes are looking good, please sign off your commit and i will merge it

franzbecker:
Okay done!

akosyakov:
@franzbecker it would be nice to add a travis badge in README

franzbecker:
@akosyakov you will need to register your company and this project with Travis first - right now Travis is not enabled for this repository. This is nothing I can do :)

Issue #11, opened 2017-07-23
Fix triggering code action from the horizontal ruler


rcjsuen:
Am I correct that `textDocument/codeActions` requests aren't happening? I'm guessing this is a Monaco problem or...?

rcjsuen:
Looks like the [marker in converter.ts](https://github.com/TypeFox/monaco-languageclient/blob/8fad2e2bc7e34e687cd010680cef4fc99cead3a2/src/converter.ts#L175-L179) has `null` for its `code` property so my language server isn't working. The diagnostics that was originally sent does have a code and it seems to be [converted correctly](https://github.com/TypeFox/monaco-languageclient/blob/8fad2e2bc7e34e687cd010680cef4fc99cead3a2/src/converter.ts#L516-L527) so I'm not sure why the code later disappears.

rcjsuen:
I opened Microsoft/vscode#31275 which explains why the `code` property is being set to `null`.

However, I still can't figure out where my code actions are for diagnostics that have a non-zero code.

akosyakov:
Do you have `glyphMargin` set to true?

rcjsuen:
@akosyakov Thanks for the tip! Setting it to `true` got the light bulb to show up. Unfortunately, clicking the light bulb doesn't do anything... :(

My language server works in VS Code and Eclipse LSP4E so I think I've implemented the protocol correctly...

akosyakov:
@svenefftinge maybe you know what can be missing on Monaco-side.

akosyakov:
@rcjsuen could be closed?

rcjsuen:
@akosyakov Not yet because clicking the light bulb doesn't do anything. :(

svenefftinge:
Have you registered the commands that are sent through the CodeActions?
Clicking the bulb or pressing (<kbd>CMD</kbd> + <kbd>.</kbd>) should trigger the `IContextMenuService` with the appropriate commands.

rcjsuen:
> Have you registered the commands that are sent through the CodeActions?

No. My language server works in VS Code without me having to explicitly register so I didn't think I had to do anything.

Using the keybinding you suggested got a menu to popup with my command. It doesn't run because of #12 but at least the menu shows up anyway.

akosyakov:
After https://github.com/TypeFox/monaco-languageclient/pull/21 gets merged there will be a light build for the code action in the example project.

[...]

PR #14, opened 2017-07-24
Convert Diagnostic's code to a string for VSCode's IMarkerData

rcjsuen:
In #11, I realized that if my Diagnostic's `code` property was set to the `0` number, Monaco would refuse to render my code actions. This is because VSCode's IMarkerData expects its `code` property to be a `string`. [`as string`](https://github.com/TypeFox/monaco-languageclient/blob/8fad2e2bc7e34e687cd010680cef4fc99cead3a2/src/converter.ts#L518) is used in converter.ts to cast the Diagnostic's `code` to a `string` but this doesn't actually do anything in the transpiled JavaScript.
```JavaScript
ProtocolToMonacoConverter.prototype.asMarker = function (diagnostic) {
    return {
        code: diagnostic.code,
        severity: this.asSeverity(diagnostic.severity),
        message: diagnostic.message,
        source: diagnostic.source,
        startLineNumber: diagnostic.range.start.line + 1,
        startColumn: diagnostic.range.start.character + 1,
        endLineNumber: diagnostic.range.end.line + 1,
        endColumn: diagnostic.range.end.character + 1
    };
};
```
The fix is to truly convert the `number` to a `string` via a call to `toString()`.

### DIFF
--- a/src/converter.ts
+++ b/src/converter.ts
@@ -515,7 +515,7 @@ export class ProtocolToMonacoConverter {
 
     asMarker(diagnostic: Diagnostic): monaco.editor.IMarkerData {
         return {
-            code: diagnostic.code as string,
+            code: typeof diagnostic.code === "number" ? diagnostic.code.toString() : diagnostic.code as string,
             severity: this.asSeverity(diagnostic.severity),
             message: diagnostic.message,
             source: diagnostic.source,
--- a/src/converter.ts
+++ b/src/converter.ts
@@ -515,7 +515,7 @@ export class ProtocolToMonacoConverter {
 
     asMarker(diagnostic: Diagnostic): monaco.editor.IMarkerData {
         return {
-            code: typeof diagnostic.code === "number" ? diagnostic.code.toString() : diagnostic.code as string,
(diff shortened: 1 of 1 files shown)


src/converter.ts
```diff
@@ -515,7 +515,7 @@ export class ProtocolToMonacoConverter {
 
     asMarker(diagnostic: Diagnostic): monaco.editor.IMarkerData {
         return {
-            code: diagnostic.code as string,
+            code: typeof diagnostic.code === "number" ? diagnostic.code.toString() : diagnostic.code as string,
```
akosyakov: Could `diagnostic.code as string` to be reduced to `diagnostic.code` now?

akosyakov:
@rcjsuen nice, thank you!

rcjsuen:
@akosyakov Thank you for being so responsive! And thank you Typefox for creating this fun project! :)

src/converter.ts
rcjsuen: You're right, it can. I just assumed that tsc would complain about this because it's a `string | number` but it didn't.

Issue #12, opened 2017-07-23
Services (Commands, Workspace) implementations for the standalone editor


rcjsuen:
Ignoring the fact that `textDocument/codeActions` doesn't seem to be working for me (#11), I can't find any references to the `workspace/executeCommand` request in the code. Is this supported?

akosyakov:
One should subclass [MonacoWorkspace](https://github.com/TypeFox/monaco-languageclient/blob/master/src/workspace.ts) and implement `applyEdit`.

If you use only one editor, you can pass your editor to your subclass and apply changes directly.

rcjsuen:
It seems like the commands that a server exposes aren't getting registered with the command registry. Every command that the server exposes in the `capabilities.executeCommandProvider.commands` array should be registered on the client with a callback to send a `workspace/executeCommand` request to the server.
```
simpleServices.ts:246 Error: command 'docker.command.convertToUppercase' not found
    at StandaloneCommandService.executeCommand (simpleServices.ts:299)
    at Action._actionCallback (quickFixWidget.ts:41)
    at Action.run (actions.ts:213)
    at ActionRunner.runAction (actions.ts:242)
    at ActionRunner.run (actions.ts:234)
    at ActionItem.BaseActionItem.onClick (actionbar.ts:165)
    at actionbar.ts:144
﻿```

**Comment by akosyakov** [comment]
You should implement [Commands](https://github.com/TypeFox/vscode-languageserver-node/blob/master/client/src/services.ts#L165) interface and provide it as part of service to the language client.

**Comment by akosyakov** [comment]
We should have a default implementation for the standalone editor that accepts a standalone editor and register commands on it.

**Comment by gatesn** [comment]
In fact, because Monaco modifies the command ID, the `MonacoLanguages` class also needs to take an editor instance and prefix the command IDs with the editor ID:

https://github.com/Microsoft/vscode/blob/875144b25bf95197aff9ed0a8229ae322701bdd1/src/vs/editor/standalone/browser/standaloneCodeEditor.ts#L170

e.g. 

```
    protected createCodeActionProvider(selector: DocumentSelector, provider: CodeActionProvider): monaco.languages.CodeActionProvider {
        return {
            provideCodeActions: (model, range, context, token) => {
                if (!this.matchModel(selector, MonacoModelIdentifier.fromModel(model))) {
                    return [];
                }
                const params = this.m2p.asCodeActionParams(model, range, context);
[...]

PR #20, opened 2017-08-10
Export services from vscode-base-languageclient

gatesn:
This lets users implement service interfaces without adding a dependency on vscode-base-languageclient. Base and connection are already supported.

Would allow users to implement the `Commands` interface as per: #12


gatesn:
Actually this creates conflicts, might be easier just to add a MonacoCommands implementation and subclass

PR #21, opened 2017-08-11
Provide support for commands

### LINKED ISSUE #11: Fix triggering code action from the horizontal ruler

### LINKED ISSUE #12: Services (Commands, Workspace) implementations for the standalone editor

gatesn:
This implements a basic applyEdit operation (more advanced ones would accept the editor and preserve / manipulate the current selections), as well as registering commands against a given editor.

Fixes #11 
Fixes #12

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -1,0 +1,16 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io). All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import { Commands, Disposable } from 'vscode-base-languageclient/lib/services';
+
+export class MonacoCommands implements Commands {
+
+    public constructor(protected readonly editor: monaco.editor.IStandaloneCodeEditor) { }
+
+    public registerCommand(command: string, callback: (...args: any[]) => any, thisArg?: any): Disposable {
+        return this.editor._commandService.addCommand(command, {
+            handler: (_accessor, ...args: any[]) => callback(...args)
+        });
+    }
+}
--- a/src/workspace.ts
+++ b/src/workspace.ts
@@ -2,9 +2,11 @@
  * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io). All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-import { MonacoToProtocolConverter } from './converter';
-import { Workspace, TextDocumentDidChangeEvent, TextDocument, Event, Emitter } from "vscode-base-languageclient/lib/services";
+import { MonacoToProtocolConverter, ProtocolToMonacoConverter } from './converter';
+import { Workspace, TextDocumentDidChangeEvent, TextDocument, Event, Emitter } from 'vscode-base-languageclient/lib/services';
+import { WorkspaceEdit } from 'vscode-base-languageclient/lib/base';
 import IModel = monaco.editor.IModel;
+import IResourceEdit = monaco.languages.IResourceEdit;
 
 export class MonacoWorkspace implements Workspace {
 
@@ -14,6 +16,7 @@ export class MonacoWorkspace implements Workspace {
     protected readonly onDidChangeTextDocumentEmitter = new Emitter<TextDocumentDidChangeEvent>();
 
     constructor(
+        protected readonly p2m: ProtocolToMonacoConverter,
         protected readonly m2p: MonacoToProtocolConverter,
         protected _rootUri: string | null = null) {
         for (const model of monaco.editor.getModels()) {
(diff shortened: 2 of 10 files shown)


src/workspace.ts
```diff
@@ -83,4 +83,36 @@ export class MonacoWorkspace implements Workspace {
         return this.onDidChangeTextDocumentEmitter.event;
     }
 
+    public applyEdit(workspaceEdit: WorkspaceEdit): Thenable<boolean> {
+        let applied = true;
+        if (workspaceEdit.documentChanges) {
```
akosyakov: please use [asWorkspaceEdit](https://github.com/TypeFox/monaco-languageclient/blob/master/src/converter.ts#L273), applying `documentChanges` is not enough

src/workspace.ts
```diff
@@ -83,4 +83,36 @@ export class MonacoWorkspace implements Workspace {
         return this.onDidChangeTextDocumentEmitter.event;
     }
 
+    public applyEdit(workspaceEdit: WorkspaceEdit): Thenable<boolean> {
```
akosyakov: please return `Promise`

gatesn:
Actually, having looked at that disgusting applyEdit function, we should iterate over the changes and then apply them all in bulk. Otherwise the undo stack looks a bit odd.

src/workspace.ts
```diff
@@ -83,4 +83,36 @@ export class MonacoWorkspace implements Workspace {
...
+        let applied = true;
+        if (workspaceEdit.documentChanges) {
+            for (const change of workspaceEdit.documentChanges) {
+                if (change.textDocument.version && change.textDocument.version >= 0) {
```
akosyakov: Are these checks necessary?

src/workspace.ts
```diff
@@ -83,4 +83,36 @@ export class MonacoWorkspace implements Workspace {
...
+                                    forceMoveMarkers: true,
+                                };
+                            }),
+                            () => [],  // Do not try and preserve editor selections.
+                        );
+                    } else {
+                        applied = false;
+                    }
+                } else {
+                    applied = false;
```
akosyakov: Is not the whole change should be applied or rejected? It seems that changes can be applied partially by this method.

src/workspace.ts
```diff
@@ -83,4 +83,36 @@ export class MonacoWorkspace implements Workspace {
...
     }
 
+    public applyEdit(workspaceEdit: WorkspaceEdit): Thenable<boolean> {
+        let applied = true;
+        if (workspaceEdit.documentChanges) {
+            for (const change of workspaceEdit.documentChanges) {
+                if (change.textDocument.version && change.textDocument.version >= 0) {
+                    const textDocument = this.documents.get(change.textDocument.uri);
+                    if (textDocument && textDocument.version === change.textDocument.version) {
+                        monaco.editor.getModel(monaco.Uri.parse(textDocument.uri)).pushEditOperations(
```
akosyakov: Should not you check that the model is defined?

src/commands.ts
```diff
@@ -0,0 +1,19 @@
+/* --------------------------------------------------------------------------------------------
+ * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io). All rights reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ * ------------------------------------------------------------------------------------------ */
+import { Commands, Disposable } from 'vscode-base-languageclient/lib/services';
+
+export class MonacoCommands implements Commands {
+
+    public constructor(private _editor: monaco.editor.IStandaloneCodeEditor) { }
```
akosyakov: `protected readonly editor`

src/commands.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    public registerCommand(command: string, callback: (...args: any[]) => any, thisArg?: any): Disposable {
+        return (this._editor as any)._commandService.addCommand(command, {
```
akosyakov: please introduce augmenting typings for internals instead of using any

src/commands.ts
```diff
@@ -0,0 +1,19 @@
...
+            handler: (id: string, ...args: any[]) => {
+                console.log("Executing command", command, id, args);
```
akosyakov: Won't it pollute console?

src/workspace.ts
```diff
@@ -83,4 +83,36 @@ export class MonacoWorkspace implements Workspace {
...
+                        applied = false;
+                    }
+                } else {
+                    applied = false;
+                }
+            }
+        } else {
+            applied = false;
+        }
+        return Promise.resolve(applied);
```
akosyakov: Could you use high-level functions as `reduce` to get rid of if-else cascade?

src/commands.ts
gatesn: Sorry, left in from testing!

src/commands.ts
gatesn: Could you explain what you mean here? Not sure I'm familiar with this technique.

akosyakov: I meant that you should add d.ts. file that augments `monaco` module add exposes `_commandService`, look at https://github.com/theia-ide/theia/blob/master/packages/monaco/src/typings/monaco/index.d.ts.

gatesn:
I've made the changes but not written an example. Any ideas on what to use as an example?!

akosyakov:
@gatesn the same here please use rebase instead of merge

akosyakov:
I like changes but there should be a proof that it is working: tests or use of it in the example.

What if the JSON server provides:
- a command that converts a property name at the position to the lower via `applyEdit`
- some way to trigger it for example via the code action

It should not be sophisticated, but only demonstrate that everything works together.

gatesn:
@akosyakov I created an example that provides a code action for any document in all positions. This action triggers a server-side command that upper cases the entire document.

akosyakov:
@gatesn I've tested this PR and it works nicely, thank you! I've opened a new PR against your, please merge it and after that, I will approve this PR.

Next time you can work in the branch directly in this repo since you have the write access instead of working on the fork. It makes easy to push new commits to PRs.

gatesn:
@akosyakov merged your changes, thanks

gatesn:
Do you think we could cut a release after this merges?

akosyakov:
@gatesn LGTM, thank you again!

akosyakov:
> Do you think we could cut a release after this merges?

yes, we can make a release. Should we migrate to [monaco 0.10.0](https://github.com/TypeFox/monaco-languageclient/issues/26) before the release?

akosyakov:
btw i wonder why github cannot recognize your account as an author, see: https://github.com/TypeFox/monaco-languageclient/commits/master

gatesn:
Probably no 0.10.0 given the breaking changes around Code actions. We might need to investigate to see what that means.

gatesn:
And it seems my git setup is different on different laptop :/

akosyakov:
@gatesn 0.2.0 is published with 0.10.0 Monaco support thanks to @rcjsuen

PR #22, opened 2017-08-11
Allow overriding the rootUri in the workspace

### LINKED ISSUE #17: Allow a rootUri to be provided to MonacoWorkspace

gatesn:
Fixes #17

### DIFF
--- a/src/workspace.ts
+++ b/src/workspace.ts
@@ -8,15 +8,14 @@ import IModel = monaco.editor.IModel;
 
 export class MonacoWorkspace implements Workspace {
 
-    protected _rootUri: string | null = null;
-
     protected readonly documents = new Map<string, TextDocument>();
     protected readonly onDidOpenTextDocumentEmitter = new Emitter<TextDocument>();
     protected readonly onDidCloseTextDocumentEmitter = new Emitter<TextDocument>();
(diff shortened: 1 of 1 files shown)


src/workspace.ts
```diff
@@ -8,26 +8,21 @@ import IModel = monaco.editor.IModel;
...
-
     protected readonly documents = new Map<string, TextDocument>();
     protected readonly onDidOpenTextDocumentEmitter = new Emitter<TextDocument>();
     protected readonly onDidCloseTextDocumentEmitter = new Emitter<TextDocument>();
     protected readonly onDidChangeTextDocumentEmitter = new Emitter<TextDocumentDidChangeEvent>();
 
     constructor(
-        protected readonly m2p: MonacoToProtocolConverter) {
+        protected readonly m2p: MonacoToProtocolConverter,
+        public readonly rootUri: string | null = null) {
```
akosyakov: It is a breaking change there are subclasses which reassign `_rootUri`. It would be better to leave `rootUri` getter and just move `protected _rootUri: string | null = null` to the constructor.

gatesn:
Fixed

akosyakov:
Thank you!

PR #25, opened 2017-08-25
fixed example + added it to the build

akosyakov:


akosyakov:
@gatesn Thanks for the review!

PR #29, opened 2017-08-30
Update Monaco dependency to 0.10.0

rcjsuen:
I bumped the versions in the `package.json` file and also fixed a compilation error caused by the removal of the [CodeAction interface](https://github.com/Microsoft/monaco-editor/commit/e5c642781a30318de8f5406fdd427f8a4acc587a).


rcjsuen:
I think the caching mechanism in Travis is causing the build to fail.

Issue #42, opened 2018-01-08
Get-rid of monaco-editor-core dependency


akosyakov:
Using monaco-editor should be enough.

PR #43, opened 2018-01-14
Fix GH-42: removed a dependency to monaco-editor-core

akosyakov:


akosyakov:
It will be a breaking change but should reduce the size of dependencies by 40MB.

akosyakov:
If nobody objects merging it tomorrow.

gatesn:
Could you describe the impact of this change? Why is it breaking?

PR #46, opened 2018-01-31
When the language server doesn't provide a resolve option, return the original codelens

Twinside:
In browser test shown that monaco still call the resolver in all cases,
so we can't use undefined here.

Thanks for the library :)

### DIFF
--- a/src/languages.ts
+++ b/src/languages.ts
@@ -271,7 +271,7 @@ export class MonacoLanguages implements Languages {
                 }
                 const protocolCodeLens = this.m2p.asCodeLens(codeLens);
                 return provider.resolveCodeLens!(protocolCodeLens, token).then(result => this.p2m.asCodeLens(result))
-            } : undefined
+            } : ((m, codeLens, t) => codeLens)
         }
     }
 


src/languages.ts
```diff
@@ -271,7 +271,7 @@ export class MonacoLanguages implements Languages {
                 }
                 const protocolCodeLens = this.m2p.asCodeLens(codeLens);
                 return provider.resolveCodeLens!(protocolCodeLens, token).then(result => this.p2m.asCodeLens(result))
-            } : undefined
+            } : ((m, codeLens, t) => codeLens)
```
akosyakov: Are you sure that it is necessary? `resolveCodeLens` is optional, I assume monaco does it automatically

Twinside: It looks like their is a bug on their side, but I'm not optimistic about the turnaround time, I'll look after making a pull request for them too, from the compiled editor.main.js (line 90163):

``` javascript
            var promises = toResolve.map(function (request, i) {
                var resolvedSymbols = new Array(request.length);
                var promises = request.map(function (request, i) {
                    return async_1.asWinJsPromise(function (token) {
                        return request.provider.resolveCodeLens(model, request.symbol, token);
                    }).then(function (symbol) {
                        resolvedSymbols[i] = symbol;
                    });
                });
                return winjs_base_1.TPromise.join(promises).then(function () {
                    lenses[i].updateCommands(resolvedSymbols);
                });
            });
```

akosyakov: looks like you are right, please file an issue for monaco-editor as well

akosyakov:
@Twinside thank you

Twinside:
Oh, you already filled a ticket: https://github.com/Microsoft/monaco-editor/issues/576

PR #49, opened 2018-02-01
fix glob-to-regexp d.ts

AlexTugarev:
Signed-off-by: Alex Tugarev <alex.tugarev@typefox.io>

### DIFF
--- a/src/languages.ts
+++ b/src/languages.ts
@@ -1,8 +1,8 @@
 /* --------------------------------------------------------------------------------------------
- * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io). All rights reserved.
+ * Copyright (c) 2018 TypeFox GmbH (http://www.typefox.io). All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-import globToRegExp from "glob-to-regexp";
+import globToRegExp = require('glob-to-regexp');
 import {
     Languages, DiagnosticCollection, CompletionItemProvider, DocumentIdentifier, HoverProvider,
     SignatureHelpProvider, DefinitionProvider, ReferenceProvider, DocumentHighlightProvider,
--- a/typings/glob-to-regexp/index.d.ts
+++ b/typings/glob-to-regexp/index.d.ts
@@ -1,7 +1,9 @@
 declare module "glob-to-regexp" {
-    export default function globToRegExp(glob: string, opts?: {
+    namespace globToRegExp {}
+    function globToRegExp(glob: string, opts?: {
         extended?: boolean;
         globstar?: boolean;
         flags?: string;
     }): RegExp;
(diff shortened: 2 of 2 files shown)


AlexTugarev:
The issue is, that `DocumentFilter`s like `{ pattern: ".*"}` cannot be used, as the type definition refers to an ES6 module, but `glob-to-regexp` is not a such. Also, the function is only called if only `pattern` is provided.

typings/glob-to-regexp/index.d.ts
```diff
@@ -1,7 +1,9 @@
 declare module "glob-to-regexp" {
```
akosyakov: Do we need to delcare module then at all? https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-function-d-ts.html looks like we should be able to declare a function and export it

src/languages.ts
```diff
@@ -1,8 +1,8 @@
 /* --------------------------------------------------------------------------------------------
- * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io). All rights reserved.
+ * Copyright (c) 2018 TypeFox GmbH (http://www.typefox.io). All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-import globToRegExp from "glob-to-regexp";
+import * as globToRegExp from 'glob-to-regexp';
```
akosyakov: `import globToRegExp = require('glob-to-regexp')` should be used

typings/glob-to-regexp/index.d.ts
AlexTugarev: module declaration is needed if `import ... from "module"` syntax should stay, cf. https://github.com/Microsoft/TypeScript/issues/5073.

src/languages.ts
AlexTugarev: should I change it?

AlexTugarev: 👍

typings/glob-to-regexp/index.d.ts
akosyakov: if we use `import x = require('module')` then it should not be necessary

AlexTugarev: well, it's still required.

> [ts]
Could not find a declaration file for module 'glob-to-regexp'. '/Users/tugarev/Projects/monaco-languageclient/node_modules/glob-to-regexp/index.js' implicitly has an 'any' type.
  Try `npm install @types/glob-to-regexp` if it exists or add a new declaration (.d.ts) file containing `declare module 'glob-to-regexp';`

AlexTugarev: I double-checked that, it's required if types should be imported, right? module-function.d.ts seems to be misleading.

akosyakov:
@AlexTugarev I've sent you an invitation, please accept it and merge the PR

PR #51, opened 2018-02-03
[example] respect a pathname in a language server URL

akosyakov:


example/src/client.ts
```diff
@@ -66,7 +67,7 @@ function createLanguageClient(connection: MessageConnection): BaseLanguageClient
 
 function createUrl(path: string): string {
     const protocol = location.protocol === 'https:' ? 'wss' : 'ws';
-    return `${protocol}://${location.host}${path}`;
+    return normalizeUrl(`${protocol}://${location.host}${location.pathname}${path}`);
```
AlexTugarev: why appending `location.pathname` now?

akosyakov: I am accessing the example server deployed on the remote machine and this machine exposing the server on the subpath instead of on the host.

PR #54, opened 2018-02-11
Add support for textDocument/documentLink and documentLink/resolve

rcjsuen:
I've updated the code to add support for `textDocument/documentLink` and `documentLink/resolve` so that Monaco will show links provided by the language server. These links will be clickable via Ctrl+Click (maybe some other key or gesture on a Mac, I don't know, sorry).

Because of the discrepancy between the LSP's `DocumentLink` and Monaco's `ILink` (see Microsoft/monaco-editor#719), I had to do some casting to get around a compiler error.


rcjsuen:
@akosyakov Thanks for reviewing. How should I merge this pull request in? Should I rebase it or should I create a merge commit?

akosyakov:
rebasing is fine

PR #63, opened 2018-03-15
Upgrade to Monaco 0.11.x

### LINKED ISSUE #11: Fix triggering code action from the horizontal ruler

rcjsuen:
There are a couple of API changes from 0.10.x to 0.11.x.

1. The new `lightbulb` was added for Microsoft/monaco-editor#574 and fixes #11.
2. The commands and code action changes were made to support `textDocument/codeAction` and `workspace/executeCommand` due to the changes from https://github.com/Microsoft/vscode/commit/d86955c7e8c332f79bd99aa1313bfff11d52f895.
3. Monaco API changes caused some refactoring in the code for supporting `textDocument/hover` and `workspace/applyEdit`.
4. I ran into some problems with union types and promises that forced me to change the signature of some of the conversion code for `CompletionItem[]` and `CompletionList`. I documented my problem in [Microsoft/TypeScript#7294](https://github.com/Microsoft/TypeScript/issues/7294#issuecomment-373303058).

### DIFF
--- a/src/converter.ts
+++ b/src/converter.ts
@@ -1,5 +1,5 @@
 /* --------------------------------------------------------------------------------------------
- * Copyright (c) 2017 TypeFox GmbH (http://www.typefox.io). All rights reserved.
+ * Copyright (c) 2017, 2018 TypeFox GmbH (http://www.typefox.io). All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 import * as is from 'vscode-base-languageclient/lib/utils/is';
@@ -13,7 +13,7 @@ import {
     Hover, SignatureHelp, SignatureInformation, ParameterInformation,
     Definition, Location, DocumentHighlight, DocumentHighlightKind,
     SymbolInformation, DocumentSymbolParams, CodeActionContext, DiagnosticSeverity,
-    Command, CodeLens, FormattingOptions, TextEdit, WorkspaceEdit, DocumentLinkParams, DocumentLink
+    Command, CodeLens, FormattingOptions, TextEdit, WorkspaceEdit, DocumentLinkParams, DocumentLink, MarkedString
 } from 'vscode-base-languageclient/lib/base';
 import IReadOnlyModel = monaco.editor.IReadOnlyModel;
 
@@ -94,7 +94,7 @@ export class MonacoToProtocolConverter {
     asCompletionItem(item: monaco.languages.CompletionItem): CompletionItem {
         const result: CompletionItem = { label: item.label };
         if (item.detail) { result.detail = item.detail; }
-        if (item.documentation) { result.documentation = item.documentation; }
+        if (item.documentation) { result.documentation = item.documentation as string; }
         if (item.filterText) { result.filterText = item.filterText; }
         this.fillPrimaryInsertText(result, item as ProtocolCompletionItem);
         // Protocol item kind is 1 based, codes item kind is zero based.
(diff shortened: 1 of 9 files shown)


rcjsuen:
I've updated the branch to depend on version `^0.11.6` which includes the fix to the IntelliSense Microsoft/monaco-editor#756 bug.

rcjsuen:
I think the union types were never supposed to work. The new `Thenable` interface in [0.11.1](https://github.com/Microsoft/monaco-editor/blob/32f54336573831e97bfd1ac07da625eb6c7133fd/monaco.d.ts#L11-L17) extends `PromiseLike` directly. The old `Thenable` interface in [0.10.1](https://github.com/Microsoft/monaco-editor/blob/2e880e008915b278482f8ee1ddbb3792e5320364/monaco.d.ts#L11-L22) was handwritten.

I can't explain why the old one works and the new one doesn't though.

akosyakov:
That's cool, I will have a look later in details.

I wonder whether we can change an example to get rid of copying monaco and packaging it with webpack instead. @rcjsuen do you think it should be a separate issue?

rcjsuen:
@akosyakov Personally I am always in favour of keeping issues separate whenever possible.

src/converter.ts
```diff
@@ -501,10 +509,30 @@ export class ProtocolToMonacoConverter {
         const contents = Array.isArray(hover.contents) ? hover.contents : [hover.contents];
         const range = this.asRange(hover.range)!;
         return {
-            contents, range
+            contents: this.asIMarkdownStrings(contents),
+            range: range
+        };
+    }
+
+    asIMarkdownString(string: MarkedString): monaco.IMarkdownString {
```
akosyakov: you can use `typeof` to avoid down casting to `any`:
```ts
asIMarkdownString(string: MarkedString): monaco.IMarkdownString {
        if (typeof string === 'string') {
            return {
                value: string
            }
        }
        const { language, value } = string;
        return {
            value: '```' + language + '\n' + value + '\n```'
        };
    }
```

src/converter.ts
```diff
@@ -531,12 +559,19 @@ export class ProtocolToMonacoConverter {
...
         }
         if (Array.isArray(result)) {
-            return result.map(item => this.asCompletionItem(item));
+            const items = result.map(item => this.asCompletionItem(item));
+            return {
+                isIncomplete: false,
+                items: items
+            }
         }
         return <monaco.languages.CompletionList>{
```
akosyakov: fyi: `<monaco.languages.CompletionList>` is redundant, if a return type only `monaco.languages.CompletionList`

src/converter.ts
```diff
@@ -531,12 +559,19 @@ export class ProtocolToMonacoConverter {
...
+                isIncomplete: false,
+                items: []
+            }
         }
         if (Array.isArray(result)) {
-            return result.map(item => this.asCompletionItem(item));
+            const items = result.map(item => this.asCompletionItem(item));
+            return {
+                isIncomplete: false,
+                items: items
```
akosyakov: fyi: could be just `items`

src/converter.ts
```diff
@@ -271,15 +271,12 @@ export class MonacoToProtocolConverter {
...
-                resource,
-                range,
-                newText: edit.newText
-            }
-        })
+    asResourceEdits(resource: monaco.Uri, edits: TextEdit[], modelVersionId?: number,): monaco.languages.ResourceTextEdit {
+        return {
+            resource: resource,
+            edits: this.asTextEdits(edits),
+            modelVersionId: modelVersionId
```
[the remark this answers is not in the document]
rcjsuen: Thanks for the tip, done.

src/converter.ts
rcjsuen: I've replaced my code with your suggested snippet, thanks!

src/converter.ts
rcjsuen: Should've removed this, thanks for catching this!

src/converter.ts
rcjsuen: Noted, thanks!

rcjsuen:
Thanks for looking at this, Anton! I appreciate it! :)

Issue #64, opened 2018-03-17
Add ESM examples


akosyakov:
Current example uses AMD to load Monaco. It would be nice to have an example demonstrating ESM integration as well: https://github.com/Microsoft/monaco-editor/blob/master/docs/integrate-esm.md

Depends on https://github.com/TypeFox/monaco-languageclient/pull/63

akosyakov:
It is turned out to be easier to switch an existing example to ESM.

akosyakov:
Although, ESM is the cleaner approach it slows down bundling by 6s on my machine.

akosyakov:
Fixed with https://github.com/TypeFox/monaco-languageclient/pull/66

Issue #39, opened 2017-11-23
Support CompletionItem.additionalTextEdits


JanKoehnlein:
`CompletionItem#additionalTextEdits` from the langauge server are currently not converted. 

I'd need that for the new Go language server to automatically add a missing import, when a function from a not yet imported package is chosen as a completion.

akosyakov:
Unfortunately, monaco does not support it. We will need to expose some internal VS code APIs to register custom completion item providers.

akosyakov:
PRs are welcomed.

It should be first fixed in vscode: https://github.com/Microsoft/monaco-editor/issues/642#issuecomment-347200856

- Monaco typings should be adjusted
- [conversions](https://github.com/Microsoft/vscode/blob/63efefc7c279e7c963d223123d7248137034252a/src/vs/editor/standalone/browser/standaloneLanguages.ts#L626) should be adjusted as well

After that, it should be fixed in this repo.

JanKoehnlein:
Now that the vscode changes are merged. What is left to do?

rcjsuen:
@JanKoehnlein If I understand correctly, I think the Monaco Editor packages need to be updated and published to npm so that it can be consumed here.

akosyakov:
Should be resolved by https://github.com/TypeFox/monaco-languageclient/pull/63

rcjsuen:
Sorry, I should've been clearer with [what I said](https://github.com/TypeFox/monaco-languageclient/issues/61#issue-305337514). This isn't fixed yet because the code is still commented out. I didn't want to just uncomment the code without verifying it.

https://github.com/TypeFox/monaco-languageclient/blob/1542fc57730755718cde3e0a0b9e257ea96cf933/src/converter.ts#L108-L114

PR #66, opened 2018-03-17
Upgrade dependencies

akosyakov:
This PR updates dependencies in the lock file and fixes the follow-up issues. It fixes as well #64 and #39.

### DIFF
--- a/src/converter.ts
+++ b/src/converter.ts
@@ -106,8 +106,9 @@ export class MonacoToProtocolConverter {
             }
         }
         if (item.sortText) { result.sortText = item.sortText; }
-        // TODO: if (item.additionalTextEdits) { result.additionalTextEdits = asTextEdits(item.additionalTextEdits); }
-        // TODO: if (item.command) { result.command = asCommand(item.command); }
+        if (item.additionalTextEdits) { result.additionalTextEdits = this.asTextEdits(item.additionalTextEdits); }
+        if (item.command) { result.command = this.asCommand(item.command); }
+        result.commitCharacters = item.commitCharacters;
         if (ProtocolCompletionItem.is(item)) {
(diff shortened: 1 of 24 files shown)


src/converter.ts
```diff
@@ -139,6 +140,24 @@ export class MonacoToProtocolConverter {
         }
     }
 
+    asTextEdit(edit: monaco.editor.ISingleEditOperation): TextEdit {
```
rcjsuen: Should name this `asISingleEditOperation` instead.

src/converter.ts
```diff
@@ -139,6 +140,24 @@ export class MonacoToProtocolConverter {
...
+        return {
+            range,
+            newText: edit.text
+        }
+    }
+
+    asTextEdits(items: monaco.editor.ISingleEditOperation[]): TextEdit[];
+    asTextEdits(items: undefined | null): undefined;
+    asTextEdits(items: monaco.editor.ISingleEditOperation[] | undefined | null): TextEdit[] | undefined;
+    asTextEdits(items: monaco.editor.ISingleEditOperation[] | undefined | null): TextEdit[] | undefined {
```
rcjsuen: Should name these `asISingleEditOperations` instead.

src/converter.ts
rcjsuen: Sorry, ignore this. I read the signatures wrong.

src/converter.ts
rcjsuen: Sorry, ignore this. I read the signatures wrong.

akosyakov:
@rcjsuen thanks for looking at it. It was not your change, dependencies have not been updated for a while. I've pushed again to address your comments.

PR #68, opened 2018-03-21
Add a CHANGELOG.md file

rcjsuen:
I think we should add a changelog to the project to help people understand what's going on between releases. I purposely didn't include the `v0.5.0-0` as I don't think that's a full release just yet...?

I do understand that maintaining this file is extra work for all of us though.


rcjsuen:
Thanks for the review as always, @akosyakov!

PR #89, opened 2018-07-24
upgrade to latest LSP

### LINKED ISSUE #24: Migrate to vscode-languageclient instead of vscode-base-languageclient
As per https://github.com/Microsoft/vscode-languageserver-node/issues/88, an abstract BaseLanguageClient has now been exposed in the library.

It looks like a pretty significant change to monaco-languageclient to implement this, however it does unfork the library so probably worth taking on.

akosyakov:
vscode compatiblity layer is introduced to make use of vscode-languageclient directly instead of vscode-base-languageclient

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,6 @@
 .vscode
 example
+examples
 node_modules
 .gitignore
 .npmignore
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ node_js:
   - "6"
   - "8"
   - "9"
-before_install:
-  - npm i -g npm@latest
 install:
   - npm install
   - cd example
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -5,13 +5,22 @@
   "requires": true,
   "dependencies": {
     "@types/body-parser": {
-      "version": "1.16.8",
-      "resolved": "https://registry.npmjs.org/@types/body-parser/-/body-parser-1.16.8.tgz",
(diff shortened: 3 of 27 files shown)


akosyakov:
Please review.
- I've finished the implementation of VS Code compatibility layer with LSP and Monaco: https://github.com/TypeFox/monaco-languageclient/pull/89/files#diff-9b485517654cf1807a16cc585e670df4. This layer only implements APIs which are used by vscode-languageclient. Resolution of VS Code module should be adjusted by clients, for example with webpack: https://github.com/TypeFox/monaco-languageclient/pull/89/files#diff-9bb3de20ef60fb0751fb0025b2cd27ea
- API-wise there are some minor breaking changes: https://github.com/TypeFox/monaco-languageclient/pull/89/files#diff-412ff84fa2527e608a8217ca35f1a30c Major change is the switch to ES6. vscode-languageclient is targeting it, so we either have to switch to ES6 or fork it again and compile down to ES5. I would like to go for ES6, since it supported by all latest browsers: https://kangax.github.io/compat-table/es6/. Clients can still transpile it to ES5 via webpack with babel-loader.
- I've also added support of latest LSP features: https://github.com/TypeFox/monaco-languageclient/blob/1a8247cbac116071ffad283cdef99982a3a4cb7f/src/services.ts#L173-L176 They are implemented via Monaco as well, except folding. For folding first we have to upgade to latest Monaco. I will do it as a separate PR.

As the next step i am going to publish `next` version and test it with Theia.

akosyakov:
@rcjsuen many thanks for trying it out

akosyakov:
We've decided to keep vscode-base-languageclient as a vscode-languageclient transpiled to es5. As soon as https://github.com/Microsoft/vscode-languageserver-node/issues/382 is resolved, we can abandon vscode-base-languageclient completely.

PR #97, opened 2018-07-31
chore(amd): amd distribution

zewa666:
the default build is done for `commonjs` only. Since consumers might using other bundlers, in my case requirejs - which requires amd bundles - it would be great to create another compile target.

typically  a structure for multi-targets would be

lib
  -> umd
  -> commonjs
  -> amd
  -> ...

Since i didn't want to override your project structure, amd here is now compiled to `dist/amd`

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
   },
   "scripts": {
     "prepare": "npm run update-vscode && npm run clean && npm run compile",
-    "compile": "tsc",
+    "compile": "tsc && tsc --outDir dist/amd --module amd",
     "watch": "tsc -w",
     "clean": "rimraf lib",
     "update-vscode": "node ./node_modules/vscode/bin/install"


akosyakov:
@zewa666 generally I am not opposed an idea. How did you know that it does what you want? I see that usually for amd everything gets bundled in one file. Does having multipile files work for you? Do you have an example?

zewa666:
Our example is hard to share as it involves large parts and various dependencies. AMD targets do not necessarily need to be bundled. Actually having them seperaten is desired in order to be able to lazy load single files as opposed to webpack bundles. We tested the PR here against our solution and it worked.

zewa666:
Ideally the resulting folder structure should be changed so that lib contains subfolders per target but this ist certainly good enough for my project to keep going. Thx

akosyakov:
Yes, but it will be y breaking change for downstream dependencies. I would prefer to stick to the current structure.

akosyakov:
@zewa666 published 0.7.1 with amd

zewa666:
Thanks for the super fast response. We're currently in the process of updating to the new version and fixing our Service registrations, so this helps to have cleaner and easier builds without the need to run a fork

PR #100, opened 2018-08-01
fix #79: upgrade to Monaco 0.13.2

### LINKED ISSUE #79: Upgrade to Monaco 0.13.x
Looks like Monaco 0.13.0 has been [released](https://github.com/Microsoft/monaco-editor/commit/6dc4fba5deb2b730d3ef8fc99262a1d5dbcb3b4d).

akosyakov:


akosyakov:
I am going to publish `next` version and test it against Theia. Any help with testing against other language servers is appreciated. 

// cc @AlexTugarev @rcjsuen if you want to test with Gitpod: https://gitpod.io#https://github.com/TypeFox/monaco-languageclient/pull/100, drop me a message if you need an access

rcjsuen:
@akosyakov I did some tests and it seems to work okay. I wasn't able to get `textDocument/documentSymbols` working but I noticed that I couldn't even get it to work in TypeScript. DO you know how I'm supposed to bring it up?

In VS Code I would tap F1 to bring up the panel drop down and then use the `@` symbol. Is it the same in Monaco?

akosyakov:
@rcjsuen I think it is `Go To Symbol...` command.

akosyakov:
I wonder whether we should do a major release again, there are some breaking changes, but we had a major just couple days ago, maybe a minor 0.7.2 woud be fine?

rcjsuen:
@akosyakov Thanks. I confirmed that it worked for me!

PR #101, opened 2018-08-01
fix: register providers only for languages matching the documentSelector

gins3000:
Previously, providers were registered for every registered language, not just for the ones matching the documentSelector array passed to the LanguageClient config.

With this fix, language-specific providers are only registered for an existing language if there is a chance of matching, i.e.:

1. the language appears in the documentSelectors list
2. there is at least one documentSelector with non-language matching criteria

This reduces the risk of a conflict with existing providers, e.g. when using monaco-languages or monaco-json.

### DIFF
--- a/src/monaco-languages.ts
+++ b/src/monaco-languages.ts
@@ -70,7 +70,9 @@ export class MonacoLanguages implements Languages {
         const completionProvider = this.createCompletionProvider(selector, provider, ...triggerCharacters);
         const providers = new DisposableCollection();
         for (const language of getLanguages()) {
-            providers.push(monaco.languages.registerCompletionItemProvider(language, completionProvider))
+            if (this.matchLanguage(selector, language)) {
+                providers.push(monaco.languages.registerCompletionItemProvider(language, completionProvider))
+            }
         };
         return providers;
     }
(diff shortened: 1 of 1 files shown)


src/monaco-languages.ts
```diff
@@ -492,4 +526,16 @@ export class MonacoLanguages implements Languages {
         return selector === model.languageId;
     }
 
+    protected matchLanguage(selector: string | DocumentFilter | DocumentSelector, languageId: string): boolean {
+        if (Array.isArray(selector)) {
+            return selector.findIndex(filter => this.matchLanguage(filter, languageId)) !== -1;
```
akosyakov: fyi: `selector.some(...)` === `selector.findIndex(...) !== -1`

akosyakov:
> This reduces the risk of a conflict with existing providers, e.g. when using monaco-languages or monaco-json.

I won't use monaco-json together with json LS. It would be confusing for a user.

akosyakov:
@gins3000 thanks for your contribution, please ⭐️ the repo 🙏 if you like it :)

src/monaco-languages.ts
gins3000: Oh I would have used "some", but I saw findIndex used in matchModel and thought it's a coding standard you want to stick to. Will keep in mind for future PRs

akosyakov: true, I will change it in: https://github.com/TypeFox/monaco-languageclient/pull/100

gins3000:
@akosyakov thanks for the thorough and fast review. I checked, and saw that, indeed, I hadn't starred the repo yet and rectified that ;)

PR #107, opened 2018-08-16
Bumping to monaco 0.14.2

Twinside:
With adaptation for the changes in the symbol API

### DIFF
--- a/src/monaco-converter.ts
+++ b/src/monaco-converter.ts
@@ -477,7 +477,7 @@ export class ProtocolToMonacoConverter {
         };
     }
 
-    asDocumentSymbolResult(values: SymbolInformation[] | DocumentSymbol[]): monaco.languages.SymbolInformation[] {
+    asDocumentSymbolResult(values: SymbolInformation[] | DocumentSymbol[]): monaco.languages.DocumentSymbol[] {
         if (DocumentSymbol.is(values[0])) {
             // FIXME when Monaco supports DocumentSymbol
             return [];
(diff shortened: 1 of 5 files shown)


src/monaco-converter.ts
```diff
@@ -477,30 +477,33 @@ export class ProtocolToMonacoConverter {
         };
     }
 
-    asDocumentSymbolResult(values: SymbolInformation[] | DocumentSymbol[]): monaco.languages.SymbolInformation[] {
+    asDocumentSymbolResult(values: SymbolInformation[] | DocumentSymbol[]): monaco.languages.DocumentSymbol[] {
         if (DocumentSymbol.is(values[0])) {
             // FIXME when Monaco supports DocumentSymbol
```
akosyakov: @Twinside this is great! Would you be able also to add missing conversion? Now it should be possible. We should do it by analogy with https://github.com/Microsoft/vscode-languageserver-node/blob/7850be5cce4ed39485150969de14947d08c47375/client/src/protocolConverter.ts#L517-L542

Twinside:
I've added the missing conversion

akosyakov:
@Twinside it looks good, thank you. I would like to test new Monaco with Theia before merging to make sure that there is no any new UI quirks.

zewa666:
@akosyakov any update on your merging tests? Looking forward to get this PR merged so we can upgrade as well. A bunch of new cool features should be available with that.

akosyakov:
Just published v0.8.0 with these changes.

PR #112, opened 2018-09-01
Ensure that SignatureHelp and SignatureHelp has non-null arrays

rcjsuen:
While the LSP allows a `SignatureHelp`'s signatures array to be optional, Monaco does not so we should ensure that it Monaco's representation of a `SignatureHelp` is properly initialized with a zero length array if the value from the language server is `null` or `undefined`.

See Microsoft/monaco-editor#1045.

### DIFF
--- a/src/monaco-converter.ts
+++ b/src/monaco-converter.ts
@@ -613,7 +613,11 @@ export class ProtocolToMonacoConverter {
     asSignatureInformation(item: SignatureInformation): monaco.languages.SignatureInformation {
         let result = <monaco.languages.SignatureInformation>{ label: item.label };
         if (item.documentation) { result.documentation = this.asDocumentation(item.documentation); }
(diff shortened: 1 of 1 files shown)


rcjsuen:
The `parameters` array in `SignatureInformation` is actually the culprit of the bug in Microsoft/monao-editor#1045.

Either way, we should fix both of these cases...

hifall:
Is this PR supposed to fix [Monaco Editor #1045](https://github.com/Microsoft/monaco-editor/issues/1045)? I have tried to upgrade to 0.8.0, but the crash issue still persists.

rcjsuen:
Hi, @hifall. It should but it sounds like I missed something.

Please open a bug and describe your setup and versions there.

hifall:
Filed as [Issue #115](https://github.com/TypeFox/monaco-languageclient/issues/115).

hifall:
Confirmed working. Thanks!

PR #119, opened 2018-09-06
fix #117: monaco-editor-core as a dev dependency

### LINKED ISSUE #117: Allow to use alternative monaco-editor-core packages
The last release of Monaco broke Theia, since they removed unused code which is used by Theia. For Theia we are going to publish an alternative package to monaco-editor-core with needed code. Here, we are going to remove runtime dependency and the client will be responsible to provide Monaco.

akosyakov:
going to try it with Theia

### DIFF
--- a/example/package-lock.json
+++ b/example/package-lock.json
@@ -3267,11 +3267,15 @@
         "minimist": "0.0.8"
       }
     },
+    "monaco-editor-core": {
+      "version": "0.14.6",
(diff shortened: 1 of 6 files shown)


AlexTugarev:
@akosyakov, would you consider to bump version for such a change?

akosyakov:
@AlexTugarev yes, it will be a new major

PR #144, opened 2019-05-07
Fix window.showMessage

jroitgrund:
I was seeing messages of the form `Request window/showMessageRequest failed with message: unexpected actions: {"title":"foo"}`, and I think I figured out the cause.

The actions for `showMessage` are not in an array of their own, they are concatenated to the arguments list: https://github.com/microsoft/vscode-languageserver-node/blob/master/client/src/client.ts#L2760

To treat them as an array, we need to use spread parameters.

### DIFF
--- a/src/vscode-api.ts
+++ b/src/vscode-api.ts
@@ -502,7 +502,7 @@ export function createVSCodeApi(servicesProvider: Services.Provider): typeof vsc
         setLanguageConfiguration: unsupported,
         onDidChangeDiagnostics: unsupported
     };
-    function showMessage(type: MessageType, arg0: any, arg1: any): Thenable<undefined | MessageActionItem> {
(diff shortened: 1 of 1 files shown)


mofux:
@akosyakov I'm also seeing this issue.  I think the fix looks valid.

akosyakov:
@mofux yes, i will apply in on the top of https://github.com/TypeFox/monaco-languageclient/pull/163

sorry, most of effort now on VS Code compatibility in Theia and monaco-languageclient did not get enough maintenance support

If you will be interested in maintaining it a bit in any kind, please let me know, see also: https://github.com/TypeFox/monaco-languageclient/issues/164

akosyakov:
See https://github.com/TypeFox/monaco-languageclient/pull/168

PR #156, opened 2019-06-14
Make Services.install be disposable

johnwiseheart:
We frequently override services, and often see the `'Language Client services has been overridden'` error message, even though it is working how we expect. 

By having `install` return a `Disposable`, it allows us to dispose of our old services first, before setting up new ones - avoiding the console error.

Other methods considered:
* `Uninstall` method
* Removing console error
* Adding parameter to `install` method to silence the console.error

Let me know if you'd prefer a different method than the current choice.

### DIFF
--- a/client/src/services.ts
+++ b/client/src/services.ts
@@ -46,11 +46,13 @@ export namespace Services {
         }
         return services;
     }
-    export function install(services: Services): void {
+    export function install(services: Services): Disposable {
         if (global[symbol]) {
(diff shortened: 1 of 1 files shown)


johnwiseheart:
Hey @akosyakov, mind taking a look at this?

akosyakov:
@johnwiseheart please rebase and merge

akosyakov:
I've rebased and tested it, everything looks fine, thank you, merging as soon as CI is green.

PR #159, opened 2019-07-13
Consider null returns for textDocument/codeAction

### LINKED ISSUE #154: Cannot read property 'map' of undefined in ProtocolToMonacoConverter.asCodeActions
When try to run version 0.9.0 with different language server (https://github.com/sumneko/lua-language-server), in browser console is error because ProtocolToMonacoConverter.asCodeActions parameter **actions** has undefined value.

rcjsuen:
As it is possible for `textDocument/codeAction` to return a `null` value, we need to guard against this and simply return an empty array if the language server sends a null back to us.

This should fix #154.

### DIFF
--- a/src/monaco-converter.ts
+++ b/src/monaco-converter.ts
@@ -440,6 +440,9 @@ export class ProtocolToMonacoConverter {
     }
 
     asCodeActions(actions: (Command | CodeAction)[]): monaco.languages.CodeAction[] {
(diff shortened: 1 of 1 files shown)


akosyakov:
@rcjsuen I've opened to handle all possible undefined values: https://github.com/TypeFox/monaco-languageclient/pull/167  Could you have a look please? 🙏

PR #166, opened 2019-08-20
Fix usage ReconnectingWebSocket example client.ts

PHameete:
Fixed the usage of the ReconnectingWebsocket in the client.ts of the example.

Was `undefined` but should be `[]` according to the documentation: https://www.npmjs.com/package/reconnecting-websocket

This was causing websocket handshake errors for Firefox and IE.


akosyakov:
@PHameete please rebase that we can try

PHameete:
@akosyakov done

PR #168, opened 2019-08-23
fix window.showMessage

akosyakov:
rebased https://github.com/TypeFox/monaco-languageclient/pull/144


akosyakov:
@mofux @jroitgrund How do you verify it?

akosyakov:
Published it under `dev` tag for testing.

PR #170, opened 2019-08-24
Update Monaco dependency in the example

rcjsuen:
`textDocument/documentLink` requests no longer work in the example because the Monaco bridge code uses 0.17.x APIs but the example still depends on 0.15.5.


akosyakov:
thank you @rcjsuen !

rcjsuen:
@akosyakov No worries, thank you for reviewing!

PR #174, opened 2019-09-02
align CompletionItemKind with Monaco 0.17.0

### LINKED ISSUE #173: CompletionItemKind converter is broken
It looks like converter from LSP's `CompletionItemKind` to Monaco's `CompletionItemKind` is broken due to this logic:

```
    asCompletionItemKind(value: CompletionItemKind): [monaco.languages.CompletionItemKind, CompletionItemKind | undefined] {
        // Protocol item kind is 1 based, codes item kind is zero based.
        if (CompletionItemKind.Text <= value && value <= CompletionItemKind.TypeParameter) {
            return [value - 1, undefined];
        };
        return [CompletionItemKind.Text, value];
    }
```

In fact, there are a lot if difference not only in zero/one based indexes.

LSP
```
namespace CompletionItemKind {
	export const Text = 1;
	export const Method = 2;
	export const Function = 3;
	export const Constructor = 4;
	export const Field = 5;
	export const Variable = 6;
	export const Class = 7;
	export const Interface = 8;
	export const Module = 9;
	export const Property = 10;
	export const Unit = 11;
	export const Value = 12;
	export const Enum = 13;
	export const Keyword = 14;
	export const Snippet = 15;
	export const Color = 16;
	export const File = 17;
	export const Reference = 18;
	export const Folder = 19;
	export const EnumMember = 20;
	export const Constant = 21;
	export const Struct = 22;
	export const Event = 23;
	export const Operator = 24;
	export const TypeParameter = 25;
}
```

Monaco
```
export enum CompletionItemKind {
        Method = 0,
        Function = 1,
        Constructor = 2,
        Field = 3,
        Variable = 4,
        Class = 5,
        Struct = 6,
        Interface = 7,
        Module = 8,
        Property = 9,
        Event = 10,
        Operator = 11,
        Unit = 12,
        Value = 13,
        Constant = 14,
        Enum = 15,
        EnumMember = 16,
        Keyword = 17,
        Text = 18,
        Color = 19,
        File = 20,
        Reference = 21,
        Customcolor = 22,
        Folder = 23,
        TypeParameter = 24,
        Snippet = 25
    }
```

This results in invalid completion item icons for some of the kinds.

akosyakov:
fix #173: Monaco has more kinds in 0.17.0 breaking a logic of adjusting position by one. This PR maps kinds directly, later, for new versions, we will need to review a mapping.

Published under `dev` tag for testing in downstream projects.

### DIFF
--- a/client/src/monaco-converter.ts
+++ b/client/src/monaco-converter.ts
@@ -168,7 +168,35 @@ export class MonacoToProtocolConverter {
         if (original !== undefined) {
             return original;
         }
-        return value + 1 as CompletionItemKind;
+        switch (value) {
+            case monaco.languages.CompletionItemKind.Method: return CompletionItemKind.Method;
+            case monaco.languages.CompletionItemKind.Function: return CompletionItemKind.Function;
+            case monaco.languages.CompletionItemKind.Constructor: return CompletionItemKind.Constructor;
+            case monaco.languages.CompletionItemKind.Field: return CompletionItemKind.Field;
+            case monaco.languages.CompletionItemKind.Variable: return CompletionItemKind.Variable;
+            case monaco.languages.CompletionItemKind.Class: return CompletionItemKind.Class;
+            case monaco.languages.CompletionItemKind.Struct: return CompletionItemKind.Struct;
+            case monaco.languages.CompletionItemKind.Interface: return CompletionItemKind.Interface;
+            case monaco.languages.CompletionItemKind.Module: return CompletionItemKind.Module;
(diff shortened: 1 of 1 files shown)


akosyakov:
@merzbird Could you give a try with `monaco-languageclient@dev`?

merzbird:
@akosyakov 
Thanks for a quick fix.
Checked locally with `0.10.1-dev.1` version, completions work fine. Do you need any other help?

akosyakov:
@merzbird nope, thanks for checking!

PR #177, opened 2019-09-23
removed vscode dependency and replaced it with @types/vscode

### LINKED ISSUE #172: Remove the VSCode dependency
As it says on the README [here](https://github.com/Microsoft/vscode-extension-vscode#readme) the vscode module has been spit into two and should now be replaced by the @types/vscode and vscode-test modules. I'm working on a PR to fix this and should have that up in a moment, however I just thought I would make an issue about it first to draw attention to it.

scarab5q:
This PR fixes #172

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
     "@types/node": "^7.0.12",
     "@types/normalize-url": "^1.9.1",
     "@types/ws": "0.0.39",
+    "@types/vscode": "1.33.0",
     "css-loader": "^0.28.11",
     "lerna": "^3.16.4",
     "monaco-editor-core": "^0.17.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -29,5 +28,8 @@
     "client",
     "example",
     "examples/browser"
-  ]
-}
\ No newline at end of file
+  ],
+  "dependencies": {
+    "@types/vscode": "1.33.0"
```
asual: Can this be a devDependency?

scarab5q: yeah just tested it with it as a dev dependency and it works so added that

akosyakov:
@asual If you are fine with changes, then just approve and merge. CI should publish next version automatically that a new one can use it. I will do a proper release when I have time to look at the project again.

PR #191, opened 2019-12-06
Implemented support for "LocationLink" in the "onDefinition" method

### LINKED ISSUE #190: Support of "LocationLink" for the "onDefinition"-Method
Hello everyone.

First of all, thank you for the work so far. The client helped very much with implementing an editor.

Background
--------------------------
I currently work on a [language-server](https://github.com/NLueg/ov-language-server) for the DSL [openVALIDATION](https://github.com/openvalidation/openvalidation).
During the implementation, I considered most of the functions of the LSP, including the "GoTo-Definition" function.

My Problem
--------------------------
During to the fact, that openVALIDATION is close to a natural language, we don't have any brackets or use the indentation for separating different elements of the grammar. That's why I need to consider the exact range of the definition of a variable.
This is also part of the LSP, but not of the monaco-languageclient.

The monaco-languageclient can only use the class `Location`, which holds the range and the name of the requested reference. The LSP provides the class `LocationLink` which additionally holds the range of the definition of the variable. 
Therefore I wanted to use this class for implementing the feature. This actually leads to an uncaught error inside the monaco-languageclient. The vscode-language-client can work fine with the same implementation.

Additional Context
--------------------------

In the following, I use a sample openVALDATION-Code which consists of a rule and a variable.
The variable is named `underage` and is located in the first two lines. Inside the second block, you can find the rule, which uses the variable in the first condition. I won't explain the rest of it, because this is are the relevant points of the following code:

image

As I said, openVALIDATION doesn't use any brackets or indentations to separate elements.
The elements of openVALIDATION are separated with blank lines.
If I am correct, the monaco-client uses the `Location` class and tries to find the whole definition with the defined brackets. 
I now try to look at the definition of the variable `underage` with the corresponding reference inside the rule. The "brackets-problem" yields to the following documentation:

image

As you cant see, just the whole code will be displayed, because monaco doesn't recognize the end of the variable.

That's why I started using the class `LocationLink` and integrated this solution inside VSCode.
This leads to the following result:

image

The exact same code yields to the following error inside the monaco-languageclient, because the `LocationLink` class is unkown to the client.

image

<br/><br/>

I tried to implement the support for `LocationLink` by myself to create a pull request, but frankly, I wasn't very successful. Sorry for the long post but I wanted to provide all relevant information.
Thank you very much for helping me.

NLueg:
Implemented the conversion for `LocationLinks` in the  `asDefinitionResult` method.
This closes #190.

During the implementation I oriented myself to the implementation in the [VSCode client](https://github.com/microsoft/vscode-languageserver-node/blob/dd78cba181136e057cfcc4bb7bc17c7c78877648/client/src/protocolConverter.ts#L553)

### DIFF
--- a/client/src/monaco-converter.ts
+++ b/client/src/monaco-converter.ts
@@ -13,7 +13,7 @@ import {
     CompletionParams, CompletionContext, CompletionTriggerKind,
     InsertTextFormat, Range, Diagnostic, CompletionItemKind,
     Hover, SignatureHelp, SignatureInformation, ParameterInformation,
-    Definition, Location, DocumentHighlight, DocumentHighlightKind,
+    Definition, DefinitionLink, Location, LocationLink, DocumentHighlight, DocumentHighlightKind,
     SymbolInformation, DocumentSymbolParams, CodeActionContext, DiagnosticSeverity,
     Command, CodeLens, FormattingOptions, TextEdit, WorkspaceEdit, DocumentLinkParams, DocumentLink,
     MarkedString, MarkupContent, ColorInformation, ColorPresentation, FoldingRange, FoldingRangeKind,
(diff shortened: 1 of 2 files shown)


CGNonofr:
Tested with angular language server which requires it, seems to work fine

akosyakov:
@RomanNikitenko could you have a look whether we need this PR as well?

PR #207, opened 2020-04-09
Bump to monaco-editor 0.20.0

CGNonofr:


dkattan:
I'd like to see this merged in, is there something I can do to help?

client/src/monaco-language-client.ts
```diff
@@ -105,6 +107,11 @@ export class MonacoLanguageClient extends BaseLanguageClient {
         }
     }
 
+    public registerProposedFeatures() {
+        this.registerFeature(new CallHierarchyFeature(this));
```
rcjsuen, line 111: Should `CallHierarchyFeature` be registered? I don't think Monaco supports this feature, does it?

CGNonofr: Indeed, an additional component will be needed to make it work I guess.

But the SemanticTokensFeature also requires some tweaks to work so... and it seems pretty harmless to register an unsupported feature (especially this one).

What do you propose? Having a method per proposed features?

rcjsuen: @CGNonofr I was more asking to make sure I hadn't missed it while looking at the code.

I don't think it's the end of the world to register it. It just needs to be made clear that this is unsupported given that Monaco does not support it.

CGNonofr: Using a comment? or by renaming the method? or just by updating the documentation?

client/package.json
```diff
@@ -18,8 +18,8 @@
     "typings": "./lib/index",
     "dependencies": {
         "glob-to-regexp": "^0.3.0",
-        "vscode-jsonrpc": "^5.0.0",
-        "vscode-languageclient": "^6.0.0",
+        "vscode-jsonrpc": "^5.0.1",
+        "vscode-languageclient": "^6.1.3",
```
RomanNikitenko, line 22: I see [here](https://github.com/microsoft/vscode/blob/94cb3bc814d6664bc701fd9d074d5260cf55505f/extensions/css-language-features/package.json#L809) and [here](https://github.com/microsoft/vscode/blob/94cb3bc814d6664bc701fd9d074d5260cf55505f/extensions/html-language-features/package.json#L204) that `6.1.0` version of `vscode-languageclient` is used for `0.20.x` version of VS Code.
I don't know if it's critical, but maybe it makes sense to keep this version close to one which is used in VS Code.

package.json
```diff
@@ -5,10 +5,10 @@
     "@types/node": "^13.1.7",
     "@types/normalize-url": "^1.9.1",
     "@types/ws": "0.0.39",
-    "@types/vscode": "1.41.0",
+    "@types/vscode": "1.43.0",
```
RomanNikitenko, line 8: should it be the same as [here](https://github.com/TypeFox/monaco-languageclient/blob/98f15c6499f23c853882c557dc18ccf7406ed1cf/client/package.json#L8)?

akosyakov:
@CGNonofr Would it be all right if @RomanNikitenko pushes to this PR to address outstanding comments and merge it? Could you make sure that the PR is opened for collaborations with maintainers: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

CGNonofr:
@akosyakov No problem for me, it's already "opened for collaborations with maintainers"!

Feel free to do whatever is needed to get it merged

RomanNikitenko:
@CGNonofr @akosyakov 
My thoughts were the following:
- I see [here](https://github.com/microsoft/vscode/blob/94cb3bc814d6664bc701fd9d074d5260cf55505f/extensions/css-language-features/package.json#L809) and [here](https://github.com/microsoft/vscode/blob/94cb3bc814d6664bc701fd9d074d5260cf55505f/extensions/html-language-features/package.json#L204) that `6.1.0` version of `vscode-languageclient` is used for `0.20.x` version of VS Code.
- `6.1.0` version of `vscode-languageclient` is based on [1.41.0 vscode ](https://github.com/microsoft/vscode-languageserver-node/blob/release/client/6.1.0/client/package.json#L8)
- so we should use the corresponding objects, for example for `Diagnostic` the difference is:  [this code](https://github.com/microsoft/vscode/blob/release/1.41/src/vs/vscode.d.ts#L4522) vs [this one](https://github.com/microsoft/vscode/blob/release/1.43/src/vs/vscode.d.ts#L4688-L4699)

Is it correct from your point of view?
Please let me know if I'm wrong and should revert my last commit.

akosyakov:
@RomanNikitenko I think it is fine to catch up with LSP. It could be also necessary for the semantic highlighting support. One has to check which minimal version vscode-languageclient is required to support it.

RomanNikitenko:
@akosyakov 
Do you mean that we can use:
- `vscode-languageclient`: 6.1.3 
- `@types/vscode`: 1.43.0
for `0.20.x` of `Monaco`?

akosyakov:
I think it should be fine.

RomanNikitenko:
> I think it should be fine.

OK, thanks, changed to 6.1.3 and 1.43.0 respectively.

akosyakov:
@RomanNikitenko You can merge it to produce `next` version for testing.

RomanNikitenko:
@CGNonofr 
thank you for your efforts related to the upgrade!

PR #240, opened 2020-12-21
Log a warning instead of an error when overriding the services

katspaugh:
If it's an error, it should throw, not just log. If it's a warning, it should log a warning. WDYT?


CGNonofr:
Indeed, it seems legit, thanks!

PR #245, opened 2021-01-04
Update documentation for starting and building

alinturbut:
I was unable to build & start as per documentation.


CGNonofr:
I just tried and it seems you're right, thanks!

PR #248, opened 2021-02-03
Bump to monaco-editor 0.22.3

CGNonofr:


CGNonofr:
@mofux Will you be able to give a look at it?

CGNonofr:
Or anyone less who is willing to help: a simple review to prevent big obvious mistakes is fine

mofux:
@CGNonofr Thanks for your work on this. I'll try to find the time reviewing the PR (and testing the changes) tomorrow.

mofux:
Unfortunately, the new approach of importing `monaco-editor-core` is not working when using the AMD loader (see https://github.com/microsoft/monaco-editor/blob/master/docs/integrate-amd.md). 

I can see the services loading and also the language servers getting initialized and RPC happening, but the monaco instance used by monaco-languageclient is not the one I'm getting with the AMD Loader. 

This makes sense, because the `import * as monaco from 'monaco-editor-core';` will force my webpack bundler to bundle the esm version of `monaco-editor-core`, which is a totally different instance from the AMD version I'm using anywhere else.

I might consider switching to the ESM version for my app - but on the other hand I think supporting the AMD version would also be nice.

Not sure how to work around this issue though, because the `monaco` instance is not exposed globally by default anymore 🤔. We might check for the existence of `window.monaco` and use that instead if it exists, but this will still carry the problem of bundling the esm `monaco-editor-core` although it will never be required.

CGNonofr:
Hi @mofux and thanks!

I'm not familliar with AMD, but wouldn't it be possible to "hack" it by defining `monaco-editor-core` by hands:

```typescript
require(['vs/editor/editor.main'], function(monaco) {
    define('monaco-editor-core', monaco)
})
```
?

mofux:
I've just migrated my app over to using the ESM version of monaco-editor-core. I still have the same problem though, the `monaco-editor-core` used by the `monaco-languageclient` is a different instance from the one my app is getting (so I guess Webpack somehow sees them as two different modules?). 

Any ideas what could be causing this? I've made sure my app is using the same version of `monaco-editor-core` that is used by `monaco-languageclient` (0.22.3).

mofux:
Ok I think I figured out what was causing `monaco-editor-core` to get bundled twice. I believe this was caused by my dev setup. The `yarn link` to the `monaco-languageclient/client` folder in the repo somehow resulted in webpack bundling the wrong monaco-editor-core. I resolved the problem by copying the compiled `monaco-languageclient/client` folder to my projects node_modules, and renaming it to `monaco-languageclient`.

So I've been finally able to test the changes from this PR and it seems to work well 👍

I'm still concerned about the imports to `monaco-editor-core` though. This will force the users of this package to use the ESM version of monaco-editor-core, and to use a bundler like `webpack` on top. Plus the alias workaround if they are using `monaco-editor` instead of `monaco-editor-core`.

I think it could be solved by passing the monaco instance directly to `MonacoServices.install()` - similar to how the command registry had to be passed. Then the rest of our code would have to work on that instance.

CGNonofr:
Ok It seems to be a good idea, I'll give it a try!

CGNonofr:
What about https://github.com/TypeFox/monaco-languageclient/pull/248/commits/5c5bd779cf411f23a7a99f327f75adfa39ecbb98 + https://github.com/TypeFox/monaco-languageclient/pull/248/commits/6ec271ab5c597d2447130eb1c57d8ec9633fa265 + https://github.com/TypeFox/monaco-languageclient/pull/248/commits/fe5adfac5d284f6b6f8efe125598d9d6ef2a6339 ?

(Btw, I fixed 2 issues I noticed during the refactor: https://github.com/TypeFox/monaco-languageclient/pull/248/commits/0bfb4e16c315e7be9c2bb813d35af7ea52af7fa6 and https://github.com/TypeFox/monaco-languageclient/pull/248/commits/ff1154af5c23fb96d759ff0e0de10b53ff91e375)

mofux:
I'm looking into it now

mofux (approved):
Works like a charm! Thanks again for putting all this effort into it. Also 👍 for the mentioned fixes, the source maps warnings were really annoying.

CGNonofr:
Thanks to you for your help!

PR #249, opened 2021-02-11
Fix vscode api

CGNonofr:
It seems that vscode-languageclient now requires the `EventEmitter` from the `vscode` api (https://github.com/microsoft/vscode-languageserver-node/blob/ee617c1f0c64a9d4ef7c6dcbb37189c7ae63ecd6/client/src/common/semanticTokens.ts#L126)


mofux (approved):
LGTM 👍

Issue #254, opened 2021-03-02
Danger with packaging vscode-jsonrpc twice - Unknown parameter structure


Ferix9288:
tl;dr Remove dependency on vscode-jsonrpc in `example` dir because it causes things to break. See this issue https://github.com/microsoft/vscode-languageserver-node/pull/719

## Context & The Problem

So after I merged in @CGNonofr 's latest PR into my custom fork, my example all of sudden broke. Specifically, I was getting this trace error:

```
console-window.ts:11 Couldn't start client Sample Language Client
showMessage @ console-window.ts:11
showMessage @ vscode-api.ts:853
(anonymous) @ client.js:2168
Promise.then (async)
start @ client.js:2164
onConnection @ client.ts:227
webSocket.onopen @ connection.ts:19
connection.js:776 Uncaught (in promise) Error: Unknown parameter structure byName
    at computeSingleParam (connection.js:776)
    at computeMessageParams (connection.js:787)
    at Object.sendRequest (connection.js:929)
    at Object.initialize (connection.ts:107)
    at MonacoLanguageClient.doInitialize (client.js:2220)
    at MonacoLanguageClient.initialize (client.js:2216)
    at client.js:2163
computeSingleParam @ connection.js:776
computeMessageParams @ connection.js:787
sendRequest @ connection.js:929
initialize @ connection.ts:107
doInitialize @ client.js:2220
initialize @ client.js:2216
(anonymous) @ client.js:2163
Promise.then (async)
start @ client.js:2164
onConnection @ client.ts:227
webSocket.onopen @ connection.ts:19
```

When I debugged into the code, the culprit is that somehow byName kind is not recognized even though the payload looks clean.

```
parameterStructures: ParameterStructures
kind: "byName"
```

```js
    function computeSingleParam(parameterStructures, param) {
        switch (parameterStructures) {
            case messages_1.ParameterStructures.auto:
                if (isNamedParam(param)) {
                    return nullToUndefined(param);
                }
                else {
                    return [undefinedToNull(param)];
                }
                break;
            case messages_1.ParameterStructures.byName:
                if (!isNamedParam(param)) {
                    throw new Error(`Recevied parameters by name but param is not an object literal.`);
                }
                return nullToUndefined(param);
            case messages_1.ParameterStructures.byPosition:
                return [undefinedToNull(param)];
            default:
                throw new Error(`Unknown parameter structure ${parameterStructures.toString()}`);
        }
    }
```

[...]

PR #255, opened 2021-03-02
Remove dependency of vscode-jsonrpc in example #254

Ferix9288:
In reference to the issue I reported in #254 

Note: the example was working perfectly fine, but for my custom fork, I was running into issues with `unknown parameter value byName`. This change helped rectify it for me and in theory should make the example solution more robust.

### DIFF
--- a/example/package.json
+++ b/example/package.json
@@ -12,7 +12,6 @@
     "reconnecting-websocket": "^3.2.2",
     "request-light": "^0.2.2",
     "vscode-json-languageservice": "^4.0.2",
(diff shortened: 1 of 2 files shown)


CGNonofr:
Thanks! I don't see how it could hurt anyway.

@mofux can you have a quick look at it and approve it?

mofux (approved):
Thanks for the PR! LGTM 👍

Ferix9288:
Awesome, thank you both @CGNonofr  and @mofux  !

PR #279, opened 2021-08-04
Fix invalid completionItem/resolve request

CGNonofr:
When `item.documentation` is not a string but a `IMarkdownString` and the `documentationFormat` is not defined, we cast the `IMarkdownString` type to a string

The consequence of it is the returned object is missing the `kind` property, which makes it not respect the LSP protocol and makes the rust language server crash

### DIFF
--- a/client/src/monaco-converter.ts
+++ b/client/src/monaco-converter.ts
@@ -219,13 +219,11 @@ export class MonacoToProtocolConverter {
         const result: CompletionItem = { label: item.label as string };
         const protocolItem = ProtocolCompletionItem.is(item) ? item : undefined;
         if (item.detail) { result.detail = item.detail; }
-        // We only send items back we created. So this can't be something else than
-        // a string right now.
         if (item.documentation) {
-            if (!protocolItem || !protocolItem.documentationFormat) {
-                result.documentation = item.documentation as string;
+            if (typeof item.documentation === 'string') {
+                result.documentation = item.documentation;
             } else {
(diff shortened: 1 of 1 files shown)


CGNonofr:
@mofux if you can have a look as usual, I would be grateful!

mofux (approved):
LGTM

CGNonofr:
Thanks, you rocks!

CGNonofr:
It seems the travis was unplugged, so no new release will be published, can anything be done?

mofux:
I guess we'll have to setup our own build pipeline. Maybe a Github Workflow will do, but I've never set one up and currently don't have much time :( Not sure if we have enough permissions for this either.

rcjsuen:
> Maybe a Github Workflow will do, but I've never set one up and currently don't have much time :( Not sure if we have enough permissions for this either.

Yeah, I think even with GitHub Actions we would need a (new?) token for publishing to npm probably since we can't just copy/paste the secret from Travis CI?

mofux:
@rcjsuen @CGNonofr 
Do you have the permission to generate a new token for publishing to npm? I unfortunately don't have permission to manage the npm package.

If not we will need the help of the original owners: @svenefftinge @akosyakov

CGNonofr:
I don't have any permission on the npm monaco-languageclient repo! Until a solution is found, I published it to @codingame/monaco-languageclient

rcjsuen:
> Do you have the permission to generate a new token for publishing to npm?

I also have no powers here unfortunately...

PR #281, opened 2021-08-05
Fix missing didOpen notification when changing a model language

CGNonofr:


mofux (approved):
Good catch. I had been running into the same issue  myself and fixed it just like you did in this PR 😉

PR #290, opened 2021-09-17
Resolve code actions

CGNonofr:
In last version of the lsp specification, the `edit` field in a CodeAction is not mandatory anymore.

In such case, the client is supposed to call the server to resolve it when it needs to.

Unfortunately, the monaco api doesn't provide the way to do it properly.

As soon as https://github.com/microsoft/monaco-editor/issues/2663 is resolved, it will need to be properly implemented, this is a temporary "hack"


CGNonofr:
@mofux still counting on you ;)

mofux (approved):
LGTM 👌

PR #297, opened 2021-10-29
Fix dropped tags in Diagnostic -> IMarkerData

rw-access:
When converting diagnostics from the language client DIagnostic from into IMarkerData, the `tag` field was dropped.

This keeps that field present in the diagnostic.

Here's confirmation of this locally with the `Unnecessary` tag on all `false` values, and no diagnostics on `true`:
Screen Shot 2021-10-29 at 4 22 43 PM


CGNonofr:
Thank you :+1:

Can you just squash your 2 commits please?

Issue #315, opened 2022-02-16
Get everything going again


kaisalmen:
Hi there, my name is Kai Salmen and I work at TypeFox since Feb 2021. We have neglected this repository far too long we are committed to contribute here again (see https://github.com/TypeFox/monaco-languageclient/issues/313#issuecomment-1041232925). I will be the first one from TypeFox jumping in again. I have not been involved here before, so I have a learning curve and of course don't know everything, but I will get there. 🙂 

I already agreed (https://github.com/TypeFox/monaco-languageclient/issues/313#issuecomment-1041284246) with @CGNonofr to open this uncluttered the dust issue. Let's gather topics here what needs to be done and then we can open other issue when needed:

Already Completed:
- Add tags from @codingame releases ✅ 
- Get the release process working again to be able to release npm packages under https://www.npmjs.com/package/monaco-languageclient (#318) ✅ 
- Update CHANGELOG.md (#318) ✅ 
- Fix build issues on Windows (#318) ✅ 
- Switch default and protected branch from `master` to `main` branch ✅ 
- Issue #331 & PR #340):
  - Some tools/libraries are now quite old (typescript for instance) ✅ 
  - The build has unexpected problems (#321) ✅  
  - Do we really need lerna? The only thing being released is the client. Would it make sense to switch to npm workspaces (this options wasn't available in 2018, of course)? ✅ 
  - Verify build on every commit: Let's use GitHub Actions. ✅ 
  - What about one **examples** folder. Move the content from `./example` to a new sub dir in `./examples`. Having two directories is confusing. ✅ 
  - #337 ✅ 
  - Do we still need the Reconnecting web socket? (https://github.com/TypeFox/monaco-languageclient/issues/313#issuecomment-1041298931) ✅ 
  - We should keep a webpack example #346
  - A new monaco-editor version was released, we should check the compatibility ✅ (#329)

Updates / General topics / Ideas to be handled in separate issues:
- Introduce semantic release (see https://github.com/TypeFox/monaco-languageclient/issues/365)
- Add eslint configuration (see https://github.com/TypeFox/monaco-languageclient/issues/366)
- Improve code where easily possible
- Should we make [@codingame/monaco-jsonrpc](https://github.com/CodinGame/monaco-jsonrpc) or [vscode-ws-jsonrpc](https://github.com/TypeFox/vscode-ws-jsonrpc) a sub-package of this repo? (see https://github.com/TypeFox/monaco-languageclient/issues/367)
[...]

PR #330, opened 2022-03-17
Make monaco workspace disposable

CGNonofr:
This include a breaking change: Monaco.install now return a Disposable instead of the services

### DIFF
--- a/packages/client/src/monaco-services.ts
+++ b/packages/client/src/monaco-services.ts
@@ -9,6 +9,7 @@ import { MonacoLanguages } from "./monaco-languages";
 import { MonacoWorkspace } from "./monaco-workspace";
 import { ConsoleWindow } from "./console-window";
 import { Services } from "./services";
+import { Disposable, DisposableCollection } from './disposable';
 
 export interface MonacoServices extends Services {
     commands: MonacoCommands
--- a/packages/client/src/monaco-services.ts
+++ b/packages/client/src/monaco-services.ts
@@ -31,10 +31,14 @@ export namespace MonacoServices {
             window: new ConsoleWindow()
         }
     }
-    export function install(_monaco: typeof monaco, options: Options = {}): MonacoServices {
+    export function install(_monaco: typeof monaco, options: Options = {}): Disposable {
+        const disposableCollection = new DisposableCollection()
+
         const services = create(_monaco, options);
-        Services.install(services);
-        return services;
+        disposableCollection.push(services.workspace)
+        disposableCollection.push(Services.install(services));
+
+        return disposableCollection;
     }
     export function get(): MonacoServices {
         return Services.get() as MonacoServices;
--- a/packages/client/src/monaco-workspace.ts
+++ b/packages/client/src/monaco-workspace.ts
@@ -4,12 +4,15 @@
  * ------------------------------------------------------------------------------------------ */
 import type * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
 import { MonacoToProtocolConverter, ProtocolToMonacoConverter } from './monaco-converter';
-import { Workspace, WorkspaceEdit, TextDocumentDidChangeEvent, Event, Emitter } from './services';
+import { Workspace, WorkspaceEdit, TextDocumentDidChangeEvent, Event, Emitter, Disposable } from './services';
 import { TextDocument } from 'vscode-languageserver-textdocument'
+import { DisposableCollection } from './disposable';
 
-export class MonacoWorkspace implements Workspace {
+export class MonacoWorkspace implements Workspace, Disposable {
 
+    protected readonly disposableCollection = new DisposableCollection()
     protected readonly documents = new Map<string, TextDocument>();
+    protected readonly documentDisposables = new Map<string, Disposable>();
     protected readonly onDidOpenTextDocumentEmitter = new Emitter<TextDocument>();
     protected readonly onDidCloseTextDocumentEmitter = new Emitter<TextDocument>();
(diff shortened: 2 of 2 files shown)


CGNonofr:
@mofux @kaisalmen currently, there no way to uninstall the services

it means we can't change the rootUri without reloading the page

kaisalmen:
Do you mind adjusting the example? Couldn't we demonstrate the functionality via a reset button: Dispose and re-init, even if no new url is used. WDYT?

CGNonofr:
Do you think it will be relevant? Nothing will happen in the interface, and we can already achieve the reset button without this feature (we don't need to uninstall the services)

kaisalmen:
The "simple" example is likely not the best place for this. The return value of `install` is just ignored there, but putting it artificially to use may not help. So, nevermind. A more advanced example or unit test could handle this, but that's not relevant here, right now.

client/src/monaco-services.ts
```diff
@@ -31,10 +31,14 @@ export namespace MonacoServices {
             window: new ConsoleWindow()
         }
     }
-    export function install(_monaco: typeof monaco, options: Options = {}): MonacoServices {
+    export function install(_monaco: typeof monaco, options: Options = {}): Disposable {
```
kaisalmen, line 34: You need to import there otherwise the code does not compile:
```javascript
import { DisposableCollection } from './disposable';
import { Disposable } from 'vscode-jsonrpc';
```

kaisalmen (changes requested):
Code does not compile without adding new imports, otherwise this is fine. Thanks.

client/src/monaco-services.ts
CGNonofr: I didn't know what this part didn't get staged before I commited => d53522ea89007ab50c0d413aa5dfc3fde7f639b0

sorry

kaisalmen: No worries, I suggested the wrong Disposable in the proposed solution 🙈

kaisalmen (approved):
LGTM

CGNonofr:
It's still not clear what is the new process here? am I supposed to merge it? Who is responsible for making a new release?

CGNonofr:
This is not really urgent, I just wanted to get rid of a workaround

Btw: What about adopting semantic release here?

kaisalmen:
> Btw: What about adopting semantic release here?

👍 enforced or relax via humans? What do you have in mind?

CGNonofr:
The more it is automatic the better I think

The most important part is using the commit message convention allowing to detect if the next version should be a patch/minor/major version

kaisalmen:
Ok from my side. Shall I merge now? It is a breaking change and should not be release before 0.19.0. I would like it to include this and at least build/dep. update. Are you ok with waiting until next release until all is done? A `0.18.x` bug-fix release could still be done from a separate branch if it really becomes necessary.

CGNonofr:
> Ok from my side. Shall I merge now? It is a breaking change and should not be release before 0.19.0. I would like it to include this and at least build/dep. update. Are you ok with waiting until next release until all is done? A `0.18.x` bug-fix release could still be done from a separate branch if it really becomes necessary.

What if I split this one into 2 PR: one with the workspace being disposable and one with the MonacoServices.install returning a disposable (which is the breaking change)? (to be able to merge the first one)

kaisalmen:
I will rebase this branch due to #340 and then we can merge it as well.

CGNonofr:
Will there be a 0.19.0 soon?

btw, shouldn't it be a 1.0.0?

kaisalmen:
> Will there be a 0.19.0 soon? 
>
>btw, shouldn't it be a 1.0.0?

Yes, soon, but maybe after Easter weekend. I already released a `0.19.0-next.0` and I want to release a `0.19.0-next.1` may be later today.
We should have a bigger version jump once eslint and semantic versions is in? Until then we could have further next releases. I already locally played around with some improvements that allow to compile the lib to es modules + we could ship bundles together in the npm package.

Btw, I just enabled https://github.com/TypeFox/monaco-languageclient/discussions We can use this to discuss things that have  a broader scope than a single issue.

CGNonofr:
ok! I'm just not a big fan of not being able to release a fix at any time if we need to

kaisalmen:
Yes, I understand that. We can do it earlier and integrate further updates/improvements later. Did you test the new build and tried the examples yourself?

CGNonofr:
Ok I will, that's not that easy. Didn't you claim the code was unchanged? ^^

kaisalmen:
Yes :-) But starting the whole example start-up changed :-)

CGNonofr:
I noticed an issue, I'll make a PR

Btw, isn't it the moment to switch to the `main` branch instead of master?

CGNonofr:
Regarding the examples: that's a shame we have to build it before starting it, can't we use ts-node instead?

kaisalmen:
Both clients examples are served by vite and don't require a build, but the node example requires one.

CGNonofr:
So what do you think about using ts-node? We just need to replace `node lib/server.js` by `node -r tsconfig-paths/register -r ts-node/register src/server.ts`

kaisalmen:
> So what do you think about using ts-node? We just need to replace node lib/server.js by node -r tsconfig-paths/register -r ts-node/register src/server.ts

Yes, go-ahead make a PR 👍

kaisalmen:
please check if README requires adjustments

kaisalmen:
We should move this discussion to #315 or https://github.com/TypeFox/monaco-languageclient/discussions/343

Issue #337, opened 2022-03-24
Does it make sense to switch to monaco-editor instead of monaco-editor-core


kaisalmen:
I was wondering if we should switch to the dependency of `monaco-editor-core` to `monaco-editor`.

Why? Because you cannot import both `monaco-editor` and `monaco-editor-core` in the same project/app. This blocks using features available from the main editor (like js support) and adding a new language + LS connection via this project, because both define the same things in global space.

As far as I understand it you can just important the "core" feature of `monaco-editor` like this:
```javascript
import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
```
and then you achieve the same as now, but you don't block the combined usage outlined above. Is this reasoning correct? WDYT? Does this have any bad implications I am overlooking?

CGNonofr:
I agree!

Btw, since we only import it for types, It won't change anything to the generated code, will it?

We can switch on monaco-editor for the example tho, we'll at least have the syntax highlighting.

CGNonofr:
The javascript won't change, but the types will changes indeed. I don't see a single reason why someone would use monaco-editor-core instead of monaco-editor

Btw importing `monaco-editor/esm/vs/editor/editor.api` instead of `monaco-editor` makes no difference here.

What is a little weird however is that monaco-editor or monaco-editor-core is not in the dependencies of monaco-languageclient (not event in peer dependencies)

kaisalmen:
Otherwise, the TypeScript compile step must go 💥 because the import does not exist?

kaisalmen:
> so ther can import it any way they want, it can be either monaco-editor or monaco-editor-core.

💡
 
> I needed to enable skipLibCheck for this reason IIRC

Ok, also good to know.

kaisalmen:
Resolved with #340

Issue #331, opened 2022-03-18
Project restructuring sandbox


kaisalmen:
This issue covers "Required next work:" from #315 
* [x] Some tools/libraries are now quite old (typescript for instance)
* [x] The build has unexpected problems (https://github.com/TypeFox/monaco-languageclient/issues/321)
* [x] Do we really need lerna? The only thing being released is the client. Would it make sense to switch to npm workspaces (this options wasn't available in 2018, of course)?
* [x] Verify build on every commit: with GitHub Actions. (https://github.com/TypeFox/monaco-languageclient/actions)
* [x] Switch dependency of `monaco-editor-core` to`monaco-editor` and make dependency obvious (see #337)

I will start a new branch and test things ([`sandbox-331`](https://github.com/TypeFox/monaco-languageclient/tree/sandbox-331)).

kaisalmen:
@CGNonofr We have GitHub Actions! 🎉

CGNonofr:
To not poluate the root folder with too many subfolders, I would have put everything inside an `example` folder

```
packages/client // npm module
packages/example/node // server part
packages/example/client // browser client
packages/example/browser // complete in browser
```
(Even If I'm not sure about the relevance of the browser example)

I see no problem from mergeing to whatever bundler you want, but is there is reason? is there an issue with webpack?

kaisalmen:
> (Even If I'm not sure about the relevance of the browser example)

Let's not waste what is already there, right?

Currently webpack build is broken on my branch after upgrading versions. vite is just fast and easier to handle, but that's personal preference, of course. We don't need production builds for the examples and here plain node (server) and vite for client/browser related stuff is faster and easier to configure.

kaisalmen:
@CGNonofr Making good progress now. client itself and examples/node and examples/client have been stripped of unneeded dependencies. Using monaco-editor now in the client example!

CGNonofr:
Will you do a single PR including all your changes?

What do you expect the work to be done?

kaisalmen:
> Will you do a single PR including all your changes?

I fear yes, because I don't know how to separate the things properly.

> What do you expect the work to be done?

Maybe today you can have a first look (draft PR).

kaisalmen:
Ok, all examples work: https://github.com/TypeFox/monaco-languageclient/tree/sandbox-331
I need to update the README and will then open a draft PR.
If you want to test it yourself (gitpod configuration is already adjusted). Best with separate checkout:
[...]

PR #340, opened 2022-04-06
Switch to npm workspaces

kaisalmen:
This implements #331 and substantially changes the repositories structure, build process and tooling. It does not change code apart from compiler error fixes required after TypeScript update:

- Switch to npm workspace and vite and thereby drop yarn, lerna and webpack
- Restructure project: All code is moved to packages directory
- Update and clean-up all dependencies
- Use monaco-editor in examples instead of monaco-editor-core #337 
- README is fully updated
- Introduce GitHub Actions
- Reconnecting websocket has been removed from the example

This PR could be best checked with a fresh checkout. Please have a look at the updated README if the instructions are good and sufficient. Please check if the example still work as expected. I think so, but I am only involved here since a couple of weeks and may have missed something or misunderstood something.


package.json
```diff
@@ -1,37 +1,30 @@
...
-    "rimraf": "^3.0.2",
-    "source-map-loader": "^0.2.3",
-    "style-loader": "^0.20.3",
-    "typescript": "^3.7.5",
-    "uglifyjs-webpack-plugin": "^1.2.4",
-    "webpack": "^3.11.0",
-    "webpack-merge": "^4.1.2",
-    "shx": "^0.3.4"
+    "editorconfig": "^0.15.3",
+    "monaco-editor": "^0.33.0",
```
CGNonofr: FYI, it seems that `^` has a different meaning on `0.x.x` versions: it will only allow patch update (`0.33.x`)

btw that's probably what we want

packages/client/src/vscode-compatibility.ts
```diff
@@ -5,4 +5,6 @@
 import { createVSCodeApi } from "./vscode-api";
 import { Services } from "./services";
 
-export = createVSCodeApi(Services.get);
\ No newline at end of file
+const vscodeCompatibility = createVSCodeApi(Services.get);
+
+export { vscodeCompatibility };
```
CGNonofr: I'm surprised it work, the previous syntax was exporting each field of the api as a different named export, now you just export a `vscodeCompatibility`

packages/example-browser/src/main.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright (c) 2018 TypeFox GmbH (http://www.typefox.io). All rights reserved.
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
-require('monaco-editor-core');
+require('monaco-editor');
```
CGNonofr: why are we using require instead of import here?

packages/example-browser/webpack.config.js
```diff
@@ -43,7 +43,7 @@ const common = {
 if (process.env['NODE_ENV'] === 'production') {
     module.exports = merge(common, {
         plugins: [
-            new UglifyJSPlugin(),
+            new terser(),
```
CGNonofr: nitpicking: class shouldn't start with a capital letter?

btw, with webpack5, i'm not sure we need to add Terser plugin by hands it we don't need to overrides options

packages/example-client/src/json-server.ts
```diff
@@ -198,7 +198,7 @@ export class JsonServer {
         const uri = URI.parse(url);
         if (uri.scheme === 'file') {
             return new Promise<string>((resolve, reject) => {
-                fs.readFile(uri.fsPath, 'UTF-8', (err, result) => {
+                fs.readFile(uri.fsPath, { encoding: 'utf8' }, (err, result) => {
```
CGNonofr: nitpicking: we can use `import fs from 'fs/promises'` to simplify this code

packages/example-client/src/json-server.ts
```diff
@@ -207,8 +207,9 @@ export class JsonServer {
             const response = await xhr({ url, followRedirects: 5 });
             return response.responseText;
         }
-        catch (error) {
-            return Promise.reject(error.responseText || getErrorStatusDescription(error.status) || error.toString());
+        catch (error: unknown) {
+            const err = error as Record<string, unknown>;
+            return Promise.reject(err.responseText || getErrorStatusDescription(err.status as number) || err.toString());
```
CGNonofr: nitpicking: May I suggest using the new `??` operator instead of `||`

README.md
```diff
@@ -1,70 +1,110 @@
...
   - [Browser](#browser)
+  - [VSCode integration](#vscode-integration)
 - [**History**](CHANGELOG.md)
 - [**License**](#license)
 
+## Introduction
+
+### Project Modernization
+
+From release 0.19.0 onward the project switched to npm workspaces. We no longer require yarn, lerna and webpack. Mostly therefore we the list of `devDependencies` is substantially shorter. All code has been moved to [./packages](./packages) directory.
```
CGNonofr: typo: we the list

CONTRIBUTING.md
```diff
@@ -36,6 +36,8 @@ A new maintainer should be listed as [a current maintainer](#current-maintainers
 
 ### Releasing
 
+**TODO: This is outdated and needs to be discussed.**
```
CGNonofr, line 39: i would LOVE this project to use semantic-release :)

README.md
```diff
@@ -24,7 +24,7 @@
 
 ### Project Modernization
 
-From release 0.19.0 onward the project switched to npm workspaces. We no longer require yarn, lerna and webpack. Mostly therefore we the list of `devDependencies` is substantially shorter. All code has been moved to [./packages](./packages) directory.
+From release 0.19.0 onward the project switched to npm workspaces. We no longer require yarn, lerna  and webpack. Mostly therefore we the list of `devDependencies` is substantially shorter. All code has been moved to [./packages](./packages) directory.
```
CGNonofr, line 27: 1 space wasn't enough? :)

CGNonofr:
There is a lot of changes so it's hard to review, there is also some comments that are not relevant anymore I think

package.json
kaisalmen: Yes, true. This needs to be adjusted.

packages/example-browser/src/main.ts
kaisalmen: This file is gone :-)

packages/example-browser/webpack.config.js
kaisalmen: All webpack config are gone.

packages/example-client/src/json-server.ts
kaisalmen: Yes, but I didn't change it in this PR. I only added:
```javascript
const err = error as Record<string, unknown>;
```

README.md
kaisalmen: Thanks

kaisalmen:
> There is a lot of changes so it's hard to review, there is also some comments that are not relevant anymore I think

My apologies. This is maily about built, tools and dependencies and not about what the code really does.
What about: Let's not do any code changes apart from needed-because-of-newer-compiler-version in this PR, but open another PR (there are at least two things you recognized, I guess there are more).

Did you have time to try it out, btw?

kaisalmen:
I added these two points to #315:
- Introduce semantic release
- Improve code where easily possible

packages/example-client/src/json-server.ts
CGNonofr: Ok let's do it afterwards then.

We should probably have a stricter eslint config that doesn't allow using `||` on anything else than boolean

kaisalmen: Yes, I eslint config is actually missing. I will put this on this list.

CGNonofr (approved):
LGTM

PR #342, opened 2022-04-08
fix: Fix vscode-compatibility using webpack

CGNonofr:


CHANGELOG.md
```diff
@@ -193,7 +193,7 @@ In order to use `vscode-languageclient` directly the compatibility layer was imp
   ```js
       resolve: {
           alias: {
-              'vscode': require.resolve('monaco-languageclient/lib/vscode-compatibility')
+              'vscode': require.resolve('monaco-languageclient/vscode-compatibility')
```
kaisalmen, line 196: Don't change an old CHANGELOG entry

kaisalmen (changes requested):
See CHANGELOG comment

CHANGELOG.md
CGNonofr: I think I replaced every occurence but you're right

btw that's the only reference to this, since the change is not needed using vite

CGNonofr: fixed, but I think we need to add something in the readme

kaisalmen: Yes, but LTGM

kaisalmen (approved):

PR #344, opened 2022-04-08
Use ts-node for example

CGNonofr:


CGNonofr:
Feel free to edit this PR @kaisalmen

kaisalmen:
The build step is still meaningful to see if there are any code errors (and here the CI is useful, because it detects mistakes). vite/esbuild does not compile ts files, ts-node breaks, I guess

CGNonofr:
> vite/esbuild does not compile ts files, ts-node breaks, I guess

I'm not sure what you mean

I removed the 2 last commits

btw: there is a `noEmit` option for typescript to just check without emitting javascript files

kaisalmen:
> there is a noEmit option

Yeah, good idea.

> I'm not sure what you mean

vite/esbuild does not invoke the tsc compiler. They just remove type information and produces javascript on the fly. That's why it is so fast, but in general before going to production you should everything compiles (no type issues, etc.)

CGNonofr:
> vite/esbuild does not invoke the tsc compiler. They just remove type information and produces javascript on the fly. That's why it is so fast, but in general before going to production you should everything compiles (no type issues, etc.)

got it!

kaisalmen (approved):

PR #354, opened 2022-05-20
Fix resolving, never overwrite by undefined

CGNonofr:


kaisalmen (approved):
Works fine. Thanks a lot.

Issue #357, opened 2022-05-22
c2p bypassConversion static method create unecessary promises


rubenfiszel:
I spent a lot of time on the debugger for #356 and I got to the bottom of it.

bypassConversion: https://github.com/TypeFox/monaco-languageclient/blob/5f056e2f658ca29b52710a27c07e95c18925607a/packages/client/src/monaco-language-client.ts#L18

in conjunction with c2p.asFormattingOptions at: https://github.com/TypeFox/monaco-languageclient/blob/d55ce36bdc4670e0a7148fa97f40dd7649273490/packages/client/src/monaco-converter.ts#L446

results in: https://github.com/microsoft/vscode-languageserver-node/blob/f97bb73dbfb920af4bc8c13ecdcdc16359cdeda6/client/src/common/formatting.ts#L77 returning a Promise instead of the value itself.

The promise does not get serialized which is the cause for #356

Issue #356, opened 2022-05-21
textDocument/formatting does not pass model options


rubenfiszel:
I might be doing something wrong but the languageClient does correctly send a `textDocument/formatting` payload to the lsp server whenever the action: `editor.getAction('editor.action.formatDocument').run()`, but the options it pass are an empty object: `{}` when I would have expected it to be the model options which have been set to `model.updateOptions({ tabSize: 4, insertSpaces: true })`. This result in deno lsp returning an error because it expects tabSize to be there.

CGNonofr:
I think you need to implement the `configurations` field in the workspace service

rubenfiszel:
@CGNonofr not sure to fully follow.

I tried inside the clientOptions: 
```
						middleware: {
							workspace: {
								configuration: (params, token, configuration) => {
									return [
										{
											tabSize: 4,
											enable: true,
											insertSpaces: true
										}
									]
								}
							}
						}
```

but that did not seem to do the trick

CGNonofr:
I think it's 'editor.tabSize'. Btw you should probably implement it in the service itself (Services.install)

rubenfiszel:
Fixed by #359

PR #358, opened 2022-05-22
Fix for #357

rubenfiszel:
@CGNonofr I tried to guess how to do an escape hatch but the whole bypassConversion check seems flaky to me. I believe it might make more sense to do a blacklist rather than the current whitelist

### DIFF
--- a/packages/client/src/monaco-language-client.ts
+++ b/packages/client/src/monaco-language-client.ts
@@ -38,7 +38,7 @@ export class MonacoLanguageClient extends BaseLanguageClient {
         });
         self._c2p = new Proxy(self._c2p, {
             get: (target: c2p.Converter, prop: string) => {
-                if (prop === 'asUri') {
+                if (prop === 'asUri' || prop === 'asFormattingOptions') {
                     return target[prop];
                 }
                 if (prop === 'asCompletionParams') {


CGNonofr:
Indeed it's not very robust! well spotted, a refactor is probably needed here

CGNonofr:
I did another PR: https://github.com/TypeFox/monaco-languageclient/pull/359

It include some typing to be sure we're only returning a promise when it is expected, what you think?

CGNonofr:
It was fixed by https://github.com/TypeFox/monaco-languageclient/pull/359, thank you very much for your investigations!

rubenfiszel:
Glad to see a clean fix! You use some advanced typing that I'm actually really curious about, will dig into it. Do you plan to do a release soon with your fix ?

kaisalmen:
Et voilà:
https://www.npmjs.com/package/monaco-languageclient/v/1.0.1

CGNonofr:
> Glad to see a clean fix! You use some advanced typing that I'm actually really curious about, will dig into it. 

There is probably a simpler way to do it tho! I don't know why `Exclude`  wasn't working

> Do you plan to do a release soon with your fix ?

I think it will come soon by @kaisalmen

rubenfiszel:
I updated and it did the trick for #356, the dev experience is actually pretty amazing now, thanks a lot!

PR #359, opened 2022-05-23
Add some typing on the code/proto converters

CGNonofr:
Another way to fix https://github.com/TypeFox/monaco-languageclient/issues/357

### DIFF
--- a/packages/client/src/converters.ts
+++ b/packages/client/src/converters.ts
@@ -1,0 +1,155 @@
+import * as p2c from 'vscode-languageclient/lib/common/protocolConverter';
+import * as c2p from 'vscode-languageclient/lib/common/codeConverter';
+import { WillSaveTextDocumentParams } from 'vscode-languageclient';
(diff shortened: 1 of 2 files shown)


CGNonofr:
@kaisalmen I think we'll quickly need a 1.0.1 :)

kaisalmen:
Yes, seems so. I didn't really have the time to look or try these changes, but the overall approach is good. It is more verbose, but it is very clear what happens 👍

kaisalmen (approved):
Haven't tested it myself, but I like the approach.

CGNonofr:
> Haven't tested it myself, but I like the approach. LGTM. Thanks 🎉

I've only tested it on the example and it didn't seem to break anything (and the original big is fixed)

Will you be able to publish it?

kaisalmen:
I want to downgrade and fix vite to 2.9.6. I have done this in two other projects, so once you merge I will push this on the main branch and then can publish.

PR #361, opened 2022-05-24
Do not include node code

CGNonofr:
Including such code can break things by using the nodejs api like `process`


kaisalmen (approved):
LGTM. Thanks

PR #370, opened 2022-05-30
Introduce global engines entries for node and npm

kaisalmen:
Shouldn't we generally stick to node LTS? npm `7.10.0` is the npm version bundled with node `16.0.0` that's why I choose that version number.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,10 @@
         "tslib": "^2.4.0",
         "typescript": "^4.6.4",
         "vite": "2.9.6"
+      },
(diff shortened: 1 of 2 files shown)


CGNonofr:
Good idea :+1:

btw, what about adding the volta config as well?

kaisalmen:
> btw, what about adding the volta config as well?

I didn't know that one. I had to look it up. It makes sense. Do you use it and do you recommend it? I am working with Windows/Linux (WSL2) hybrid machines, so I am always interested in things that work in both worlds.

PR #372, opened 2022-06-03
Clean unregistered featured from registration requests

CGNonofr:
For some reasons, if the server sends a capability registration batch to the client and one feature is not registered on the client, all registration after it in the list won't be registered (see https://github.com/microsoft/vscode-languageserver-node/blob/a6f0d22d2c6842337ad135c13d2f945e53b1a6e4/client/src/common/client.ts#L1741)

Do you have a better idea?


kaisalmen (approved):
@CGNonofr I don't have better idea. How/when the exception is thrown must be done differently (e.g. result object / summary), then there is another way.

CGNonofr:
The only other way is to create a PR on the vscode-languageserver-node itself, I'll try to have a look :+1:

PR #386, opened 2022-06-20
Add example with client and server both running in the browser

### LINKED ISSUE #368: Should we change examples/browser (make use of vscode-languageclient/browser)?
This is an idea from https://github.com/TypeFox/monaco-languageclient/issues/315

Is it possible to integrated `vscode-languageclient/browser` here and change the browser example to make use of it?
https://code.visualstudio.com/api/extension-guides/web-extensions#language-server-protocol-in-web-extensions
https://github.com/microsoft/vscode-extension-samples/tree/main/lsp-web-extension-sample

kaisalmen:
This resolves #368 and implements client and server in the browser. Server runs in a worker and they use `vscode-languageserver-protocol/browser` for communication.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## [2.0.0] - Not yet release
+
+* TBA
+
 ## [1.1.0] - 2022-06-08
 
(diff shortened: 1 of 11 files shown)


CGNonofr (approved):
LGTM, nice addition!

Issue #388, opened 2022-06-21
No `cjs` fragments to export


char0n:
Some packages like [client](https://github.com/TypeFox/monaco-languageclient/blob/main/packages/client/package.json#L28) exports `cjs` fragments. `cjs` fragments are not part of npm distribution anymore. Exports needs to go away. Working on PR.

![image](https://user-images.githubusercontent.com/193286/174875542-416b73ac-6a65-40bf-adc7-576d13531a65.png)

kaisalmen:
Hi @char0n sorry, my mistake. The files are still there, but not packaged. I will provide a 2.0.1 soon.

kaisalmen:
New release is available:
https://www.npmjs.com/package/monaco-languageclient/v/2.0.1

kaisalmen:
@char0n Thank you

CGNonofr:
@kaisalmen it seems that @CodinGame/monaco-vscode-api can only work in esm mode, since it imports code from monaco-editor which is in esm to prevent bundling too many things vrom vscode. The library still exports an amd and cjs versions but they still import esm from monaco-editor, I don't think it makes much sense.

The amd version of monaco editor is tweaked and use a loader which prevent access to some internal tools that monaco-vscode-api requires (I don't exactly know how it's built but the files are different)

I plan to keep only the esm output of the lib (is there any other solution?) which makes any other output type of monaco-languageclient pointless

kaisalmen:
@char0n this is resolved and we kept the cjs exports for the `monaco-converters`

PR #389, opened 2022-06-21
chore(package): remove obsolete exports entries

char0n:
Refs #388


kaisalmen:
2.0.1 contains the missing files again, so the exports are still valid.

PR #390, opened 2022-06-22
Align all tsconfigs and vscode-ws-jsonrpc provides esm/cjs

kaisalmen:
Added `typesVersions` to monaco-languageclient and vscode-ws-jsonrpc, so exports can be properly imported with TS:

I did not understand why the "exports" lead to import problems in typescript (node example) until I found this:
https://github.com/teppeis/typescript-subpath-exports-workaround

`vscode-ws-jsonrpc` gets a 1.0.1 release once this is merged.
and it makes sense to release a 2.0.2 of `monaco-languageclient`, because of the  `typesVersions`. Do you have other things that should be fixed as well?

### DIFF
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this npm module are documented in this file.
 
+## [2.0.2] - 2022-06-22
+
+* Align tsconfig and use `typesVersions` in **package.json** for proper TypeScript import support (used in node example)
+
 ## [2.0.1] - 2022-06-21
 
 * No cjs fragments to export #[388](https://github.com/TypeFox/monaco-languageclient/issues/388)
@@ -12,6 +16,11 @@ All notable changes to this npm module are documented in this file.
 * Merge monaco-jsonrpc and vscode-ws-jsonrpc into new sub package #[383](https://github.com/TypeFox/monaco-languageclient/pull/383)
 * Integrate eslint with eslint-config-standard #[385](https://github.com/TypeFox/monaco-languageclient/pull/385)
 * Add example with client and server both running in the browser #[386](https://github.com/TypeFox/monaco-languageclient/pull/386)
+
+## BREAKING CHANGES
+
+* If you use Webpack or vite for bundling, you have to remove the `vscode` alias entry from the configuration
+* If you customized monaco services, then you have to add them
 
 ## [1.1.0] - 2022-06-08
 
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -2,117 +2,127 @@
 
 All notable changes to this npm module are documented in this file.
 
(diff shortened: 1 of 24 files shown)


CGNonofr:
> Do you have other things that should be fixed as well?

I admit I didn't experience any issue

packages/client/CHANGELOG.md
```diff
@@ -13,6 +17,11 @@ All notable changes to this npm module are documented in this file.
 * Integrate eslint with eslint-config-standard #[385](https://github.com/TypeFox/monaco-languageclient/pull/385)
 * Add example with client and server both running in the browser #[386](https://github.com/TypeFox/monaco-languageclient/pull/386)
 
+## BREAKING CHANGES
+
+* If you use Webpack or vite for bundling, you have to remove the `vscode` alias entry from the configuration
+* If you customized monaco services, then you have to add them
```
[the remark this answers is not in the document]
kaisalmen: "adjust them"

Do we have an example for that inside this repo

packages/examples/node/src/server.ts
```diff
@@ -7,7 +7,7 @@ import * as http from 'http';
 import * as url from 'url';
 import * as net from 'net';
 import express from 'express';
-import * as rpc from 'vscode-ws-jsonrpc';
+import * as rpc from 'vscode-ws-jsonrpc/cjs';
```
kaisalmen, line 10: @CGNonofr this did not work before I added the `typedVersions`. tsc/vscode complain it is not defined

kaisalmen: I had the same thought. This is a node application and when I switched to modules/esm then I ran into import errors beyond `vscode-ws-jsonrpc`. Before everything was commonjs and it worked. I ensure it works the same way now, too. Here no bundler is involved (vite or webpack could handle this problem)

I need to look into module only node/ts, but I just wanted to make it work again for now and provide people with options.

CGNonofr: btw, I use `node --loader ts-node/esm` instead of `node -r ts-node/register` in my projects, maybe it helps?

kaisalmen: I can give it a try

kaisalmen:
Shall I remove the AMD and CJS build from `monaco-languageclient` with this PR as well. We only keep `monaco-converter` and `monaco-converter/cjs` for BW-compatibility reasons, then?

CGNonofr:
> Shall I remove the AMD and CJS build from `monaco-languageclient` with this PR as well. We only keep `monaco-converter` and `monaco-converter/cjs` for BW-compatibility reasons, then?

Yep! I wonder why anything else than ESM would be useful in 2022. And then I noticed vscode-languageclient is still in CJS only :disappointed:

kaisalmen:
@CGNonofr I update the CHANGELOGs (sorry for the dash changes).

Do you think the BREAKING CHANGE section is now better (in both packages)?

PR #399, opened 2022-07-08
Update vscode api

CGNonofr:
It may look like a downgrade, but don't get fooled!

At first, I was trying to implement the latest available vscode api, but it was a mistake.

I'm now implementing the vscode api corresponding to the last released monaco version (which is quite old, > 3 months) so the 1.67 (instead of the 1.68)

It will be a little weird that the last published version is not the hightest number until a new monaco editor version will be released and I will be able to release a v1.69 of monaco-vscode-api.

The new version of monaco-vscode-api is using more code from vscode itself, and less copy/pastes, which I assumed will be more robust. It also implements some endpoints that weren't implemented yet.

It also exports 2 new modules: `service-override/messages` and `service-override/modelEditor` which allow to register vscode services instead of default monaco standalone services.
- `messages` are notification and dialogs, see 3da5185d100b21afa51506efeaf496dac208c158
- `modelEditor` are for model resolvers and editor management (advanced usage)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -351,14 +351,14 @@
       }
     },
     "node_modules/@typescript-eslint/eslint-plugin": {
-      "version": "5.30.5",
-      "resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-5.30.5.tgz",
-      "integrity": "sha512-lftkqRoBvc28VFXEoRgyZuztyVUQ04JvUnATSPtIRFAccbXTWL6DEtXGYMcbg998kXw1NLUJm7rTQ9eUt+q6Ig==",
+      "version": "5.30.7",
+      "resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-5.30.7.tgz",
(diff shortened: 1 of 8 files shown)


CGNonofr:
@kaisalmen I also updated vscode-languageclient and vscode-jsonrpc

Will you be able to release it soon?

kaisalmen:
@CGNonofr this afternoon, I think

kaisalmen:
Do you want to merge the branch yourself? I am preparing the CHANGELOG. New version should be 2.1.0 due to the "downgrade".

CGNonofr:
I just updated monaco-vscode-api to the last version if you want to retest one last time

As soon as it's ok to you feel free to merge it!

kaisalmen:
Ok, I guess we need 1.67.20 once you get the confirmation that your fix works. I locally updated to vite 3.0.2 and tested your changes. It looks good. I will take the opportunity to update vite along after we merged this PR and then perform the releases.

CGNonofr:
> Ok, I guess we need 1.67.20

Not really, the broken feature was a configuration service override which is not mandatory, I can't change the dependency version but it won't change anything since user will already get the last version

PR #420, opened 2022-09-14
Issue 419: Release 4.0.1

kaisalmen:
Release 4.0.1 brings back monaco-converter one last time 🙂
I added a deprecation message directly in `monaco-converter.ts`.

### DIFF
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -2,11 +2,15 @@
 
 All notable changes to this npm module are documented in this file.
 
+## [4.0.1] - 2022-09-14
+
+- monaco-converter removal [#419](https://github.com/TypeFox/monaco-languageclient/issues/419): `monaco-converter` has been re-introduced to help dependent projects adapt
+
 ## [4.0.0] - 2022-09-08
 
 - **BREAKING CHANGE**: Transform to package of type module and switch to TypeScript ECMAScript Module Support in Node.js.
 - **BREAKING CHANGE**: All cjs exports have been removed.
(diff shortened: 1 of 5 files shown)


packages/client/CHANGELOG.md
```diff
@@ -2,11 +2,15 @@
 
 All notable changes to this npm module are documented in this file.
 
+## [4.0.1] - 2022-09-14
+
+- monaco-converter removal [#419](https://github.com/TypeFox/monaco-languageclient/issues/419): `monaco-converter` has been re-introduced to help dependent projects adapt
```
CGNonofr, line 7: `removal [...] has been re-introduced` seems a little awkward to me, I'm not sure if it was removed and put back

CGNonofr (commented):
I have the feeling it creates noises and I'm not sure to understand the exact point of it, but :shrug: LGTM

kaisalmen:
> I have the feeling it creates noises and I'm not sure to understand the exact point of it, but 🤷 LGTM

Yes, I should have not directly reacted to hit. I wanted to help, because I removed the stuff without warning. Although this is not really true, because there were deprecation warnings in place already.

Therefore, we should not merge this and actually drop the idea to bring `monaco-converter` back.

I have looked at the swagger-editor project and they actually don't need `monaco-languageclient`, they can directly use rely on `monaco-vscode-api`.

4.0.1-next.0 is released, but it does no harm as it is tagged as **next**.

CGNonofr:
It was a major version, major versions break things.

Also there is an alternative, and it is not that hard to use (just switch from monaco.XXX to vscode.XXX)

PR #435, opened 2022-11-28
Library version compatibility table

### LINKED ISSUE #433: Specify monaco-editor compatibility
We should add a table to the README listing the compatible versions of:
- **monaco-editor**
- **monaco-vscode-api**
- **monaco-languageclient**

We should investigate if it is possible to "enforce" the compatibility check with a proper `peerDependencies` definitions in the **package.json**.

kaisalmen:
This resolves #433


kaisalmen:
I forgot to check the `peerDependencies`. I will do this now, too.

kaisalmen:
@CGNonofr I added the following peerDependencies:
```
"peerDependencies": {
    "monaco-editor": ">= 0.34.0 < 0.35.0",
    "vscode": ">= npm:@codingame/monaco-vscode-api@1.69.0 < npm:@codingame/monaco-vscode-api@1.70.0"
}
```
Do, you think they are to tight?

Interestingly, I was only to verify that the definitions are obeyed via a separate project and a next version available on npmjs. My first naive approach was to define mismatching dependencies in one of the examples, but that had no effect. Only after publishing a next version and with an independent dummy project, I was able to verify that the peerDeps are obeyed. E.g. monaco-editor 0.33.0 leads to install failure. 0.34.0 and 0.34.1 are fine.

CGNonofr:
Can you explain to me again the difference between `>= 0.34.0 < 0.35.0` and `~0.34.0`?

> Do, you think they are to tight?

I don't know... the constraints should be on monaco-vscode-api, not here, that's a shame that npm is not able to handle it properly.

So being very precise here can't hurt :shrug:

kaisalmen:
> Can you explain to me again the difference between >= 0.34.0 < 0.35.0 and ~0.34.0?

Don't know 😆 I looked it up. They are equal! The notation I choose is more lengthy, but you know what's the deal 😉 If you are ok with it, I like to keep it.

> don't know... the constraints should be on monaco-vscode-api,

True, but not really super bad, because people can see it (direct definition in package.json and ref in README)

>

kaisalmen:
>  the constraints should be on monaco-vscode-api,

You could add those ... should I open a PR?

CGNonofr:
> > the constraints should be on monaco-vscode-api,
> 
> You could add those ... should I open a PR?

Of course, feel free. I don't know why we use `^0.34.0` since it means any version of monaco editor (when the first number is a 0, everything is shifted to the left :shrug:)

xhalo32:
I think `">= npm:@codingame/monaco-vscode-api@1.69.0 < npm:@codingame/monaco-vscode-api@1.70.0` is not proper syntax at all. NPM thinks it's a local file dependency, as can be seen in `package-lock.json`:

```json
  "dependencies": {
    "vscode": {
      "version": "file:>= npm:@codingame/monaco-vscode-api@1.69.0 < npm:@codingame/monaco-vscode-api@1.70.0"
    }
  }
```

Also running `npm ls vscode` complains:

```
vscode@ invalid: ">= npm:@codingame/monaco-vscode-api@1.69.0 < npm:@codingame/monaco-vscode-api@1.70.0" from the root project (>= npm:@codingame/monaco-vscode-api@1.69.0 < npm:@codingame/monaco-vscode-api@1.70.0)
```

The correct syntax would have been

```json
    "vscode": "npm:@codingame/monaco-vscode-api@ >= 1.69.0 < 1.70.0"
```

> My Node.js version is `v18.20.2` and NPM is `10.5.0`

This issue went unnoticed because the `vscode` dependency was also in the `dependencies` section (which I guess caused NPM to ignore the `peerDependecy`) as follows:

```json
"vscode": "npm:@codingame/monaco-vscode-api@1.69.12",
```

Would it be possible to release version 4.0.4 with the correct syntax?

> Backstory: I was running into this while trying to bundle a project with `monaco-languageclient 4.0.3` as a dependency using [node2nix](https://github.com/svanderburg/node2nix).

xhalo32:
I committed the fixes [to my fork](https://github.com/xhalo32/monaco-languageclient/tree/fix-vscode-peerdep), feel free to merge them into your tree and publish on NPM.

Best regards,
Niklas Halonen

kaisalmen:
Hi @xhalo32 we no longer use this syntax.

xhalo32:
Hi @kaisalmen 

Yes, I know, but every downstream dependency that uses 4.0.3 "inherits" this bug. What I'm suggesting is [backporting](https://en.wikipedia.org/wiki/Backporting) the fix to the version syntax so that the fix is available for those who can't upgrade to version 5 or higher (like me :smile: ).

kaisalmen:
> What I'm suggesting is [backporting](https://en.wikipedia.org/wiki/Backporting) the fix to the version syntax so that the fix is available for those who can't upgrade to version 5 or higher (like me 😄 ).

@xhalo32 currently there is no intention to maintain multiple versions. Feel free to fork the old version.

Issue #412, opened 2022-08-23
React Compatibility 


wudstrand:
I am able to get the client example to connect with my python LSP server, but when I try to use the library in my react app it isn't working. I was wondering if we could get some docs on how to integrate with react, or if anyone knew of any guides on how to connect a custom LSP server to the `react-monaco-editor`.

kaisalmen:
Hi @wudstrand 

> I try to use the library in my react app it isn't working

Can you be more specific about that? What does not work? We don't have a react specific example, only the webpack related one to show how to use it.

CGNonofr:
Are you using @monaco-editor/react or react-monaco-editor? `@monaco-editor/react` won't work as it doesn't use the ESM version

What is the result of a `npm list monaco-editor`?

wudstrand:
I am using the `@monaco-editor/react` at the moment so I guess I need to transition over to `react-monaco-editor` library. Do I need to use web pack with this component? Here is a copy of my current impl of the editor in react. If you could let me know if I am on the right track that would be great.

```typescript
import React, {useState} from 'react';
import Editor, {EditorProps, Monaco} from "@monaco-editor/react";
import {toSocket, WebSocketMessageReader, WebSocketMessageWriter} from "vscode-ws-jsonrpc";
import {
    CloseAction,
    ErrorAction,
    MessageTransports,
    MonacoLanguageClient,
} from "monaco-languageclient";
import normalizeUrl from 'normalize-url';



const ReconnectingWebSocket = require('reconnecting-websocket');


interface Props extends EditorProps {
    enableEdit?: boolean
}

const LSPConnectedEditor: React.FC<Props> = ({enableEdit, ...other}) => {
    const [editor, setEditor] = useState<any | null>(null)

    function createLanguageClient (transports: MessageTransports): MonacoLanguageClient {
        return new MonacoLanguageClient({
            name: 'Sample Language Client',
            clientOptions: {
                // use a language id as a document selector
                documentSelector: ['python'],
                // disable the default error handler
                errorHandler: {
                    error: () => ({ action: ErrorAction.Continue }),
                    closed: () => ({ action: CloseAction.DoNotRestart })
                }
            },
            // create a language client connection from the JSON RPC connection on demand
            connectionProvider: {
                get: () => {
                    return Promise.resolve(transports);
                }
            }
        });
    }

    function createWebsocket(socketUrl: string) {
        const socketOptions = {
            maxReconnectionDelay: 10000,
[...]

PR #436, opened 2022-11-29
React client example

### LINKED ISSUE #415: Add a React example
Add an React usage example to the examples

kaisalmen:
This resolves #415 and #412.

This realizes standard client to node LSP server as React functional component.


packages/examples/server/src/server.ts
```diff
@@ -40,6 +40,7 @@ server.on('upgrade', (request: http.IncomingMessage, socket: net.Socket, head: B
         wss.handleUpgrade(request, socket, head, webSocket => {
             const socket: rpc.IWebSocket = {
                 send: content => webSocket.send(content, error => {
+                    console.log(content);
```
CGNonofr, line 43: is it there on purpose?

packages/examples/react/src/app.tsx
```diff
@@ -77,82 +120,27 @@ export class ReactMonacoEditor extends React.Component<EditorProps> {
...
 
-            this.createWebSocket();
+            createWebSocket(url);
 
-            this.resizeEvent = () => this.editor?.layout();
-            window.addEventListener('resize', this.resizeEvent);
-        }
-    }
+            const resizeEvent = () => editorRef.current!.layout();
+            window.addEventListener('resize', resizeEvent);
```
CGNonofr, line 126: What is the issue with `automaticLayout`?

packages/examples/react/src/app.tsx
```diff
@@ -77,82 +120,27 @@ export class ReactMonacoEditor extends React.Component<EditorProps> {
...
 
-    componentDidUpdate(_prevProps: EditorProps) {
-        const { text } = this.props;
-        const model = this.editor.getModel();
-        if (model && text !== model.getValue() && text) {
-            model.setValue(text);
+            console.log('React component initialized');
```
CGNonofr, line 128: Is it there on purpose?

packages/examples/react/src/app.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { StandaloneServices } from 'vscode/services';
+import getMessageServiceOverride from 'vscode/service-override/messages';
+
+StandaloneServices.initialize({
+    ...getMessageServiceOverride(document.body)
+});
+buildWorkerDefinition('dist', new URL('', window.location.href).href, false);
+
+export interface EditorProps {
+    text: string;
```
CGNonofr, line 38: `text` doesn't seem like a very good name, because it's only used when the component is mounted

what about `defaultText`? or `defaultCode`?

packages/examples/server/src/server.ts
kaisalmen: Yes, I put this in to see if content is sent and I thought it may be helpful, because it is an example, but we can remove it again,

packages/examples/react/src/app.tsx
kaisalmen: Yes, true, I will change to `code` . As I understand it the props could be used on update (not done in this example), so no need to have the `default`prefix.

packages/examples/react/src/app.tsx
kaisalmen: Both log statement can be removed.

packages/examples/react/src/app.tsx
kaisalmen: Oh boy, good point. 🙈 I wasn't aware this is possible or I seem to have forgotten about. Now you asked the question, I tested it and it works nicely.

CGNonofr: I think on earlier version of monaco it wasn't great as they were polling, but for some versions now they use a ResizeObserver

packages/examples/react/src/app.tsx
CGNonofr: I would expect a `code` prop to be used when updated, since it's not, it's confusing. Some big libraries in react use the `value`/`defaultValue` convention (like https://github.com/suren-atoyan/monaco-react#usage)

packages/examples/react/src/app.tsx
kaisalmen: 👍  Good to know. I am missing some history here.

kaisalmen:
I implemented all your review comments (also the latest ones) and renamed the example from react to react-client

kaisalmen:
Build is fixed. Btw, I updated the tsc target to ES2022 on main. I found some inconsistencies in the build configuration while working on react and angular examples

CGNonofr (approved):

kaisalmen:
Thanks for the feedback.

PR #438, opened 2022-11-30
Angular client example

### LINKED ISSUE #414: Add an angular example
Add an angular usage example to the examples

kaisalmen:
This resolves #414 

It is working already using ng dev server. codicon is not loaded properly and editor worker is not yet working


index.html
```diff
@@ -20,7 +20,8 @@ <h1>Examples</h1>
         <br><br>
         <a href="packages/examples/react/index.html">React Client Example</a>
         <br><br>
-        <a href="packages/examples/angular/index.html">Angular Client Example</a>
+        Please start <b>npm run start:example:angular</b> beforehand:<br>
```
CGNonofr, line 23: no big deal, but better use `<code>npm run start:example:angular</code>`

CGNonofr (commented):
LGTM

index.html
kaisalmen: Good idea, thanks. I adjusted that ⬇️

kaisalmen:
`codicon.ttf` loading can be painful. Here the problem is that angular is a workspace project and ttf loading from inside monaco is forbidden as node_modules is outside the root of the angular project.

It is now copied locally and loaded via font-face style instruction. There is still an error in the console, but it works. I have tried multiple webpack config adjustments that usually work, but the angular being a sub-package is not working well.

kaisalmen:
We can improve this later (if need and ideas are there)

PR #443, opened 2023-01-02
Close websocket

w3ichen:
**Issue**
When I implemented the current react-client example on a long-running server instance, I saw exploding memory usage but no CPU usage. The memory hung for a long time but was never released. I checked my console network tab, and the connections were never being closed.  

**This pull request**
I believe there need to be calls made to close the web sockets, or else there will be a build up of memory on the server.
I added calls to close the web socket on component unmount and page exit/reload

### DIFF
--- a/packages/examples/react-client/src/app.tsx
+++ b/packages/examples/react-client/src/app.tsx
@@ -132,11 +132,11 @@ export const ReactMonacoEditor: React.FC<EditorProps> = ({
 
         window.onbeforeunload = () => {
           // On page reload/exit, close web socket connection
-          !!lspWebSocket && lspWebSocket.close();
+          lspWebSocket?.close();
         };
         return () => {
             // On component unmount, close web socket connection
-            !!lspWebSocket && lspWebSocket.close();
+            lspWebSocket?.close();
         };
     }, []);
 
--- a/packages/examples/react-client/src/app.tsx
+++ b/packages/examples/react-client/src/app.tsx
@@ -60,6 +60,7 @@ function createWebSocket(url: string) {
         languageClient.start();
         reader.onClose(() => languageClient.stop());
     };
+    return webSocket;
 }
 
 function createLanguageClient(transports: MessageTransports): MonacoLanguageClient {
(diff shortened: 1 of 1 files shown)


kaisalmen (approved):
LGTM! Thank you 🎉

packages/examples/react-client/src/app.tsx
```diff
@@ -121,14 +123,21 @@ export const ReactMonacoEditor: React.FC<EditorProps> = ({
...
 
             return () => {
                 editorRef.current!.dispose();
             };
         }
 
-        return () => {};
+        window.onbeforeunload = () => {
+          // On page reload/exit, close web socket connection
+          !!LSPWebSocket && LSPWebSocket.close();
```
CGNonofr, line 135: Can we just use a modern regular syntax instead of this obfuscated code?

```typescript
if (LSPWebSocket != null) {
   LSPWebSocket.close();
}
```
Also the variable probably shouldn't start with a capital letter

kaisalmen: Oh, yeah, good point. The variable should be `lspWebSocket`. I don't mind the syntax, but yes, simpler is better.

w3ichen: Thanks, I've renamed it to `lspWebSocket`

w3ichen: @CGNonofr I did `!!lspWebSocket && lspWebSocket.close();` so that it only runs if `lspWebSocket` is defined.
Your approach works too. I just did it this way because, for my use cases, `lspWebSocket` will not always be defined, such as when the languageclient is disabled

kaisalmen (changes requested):
@CGNonofr is right
`lspWebSocket?.close();` and `!!lspWebSocket && lspWebSocket.close();` achieve the same, but the first statement is more consistent with the rest of the code here and also easier to understand IMO. Sorry for the confusion I created.

w3ichen:
@kaisalmen @CGNonofr I made the change to `lspWebSocket?.close();`!

kaisalmen (approved):
LGTM! Thank you 🎉

Issue #445, opened 2023-01-05
`rpc.isNotificationMessage()` missing


arnoson:
I'm trying to recreate the server side connection forwarding example in the [vscode-ws-jsonrpc readme](https://github.com/TypeFox/monaco-languageclient/tree/main/packages/vscode-ws-jsonrpc):

```ts
import * as rpc from 'vscode-ws-jsonrpc';
import * as server from 'vscode-ws-jsonrpc/server';

const socket: rpc.IWebSocket; // open the web socket
const reader = new rpc.WebSocketMessageReader(socket);
const writer = new rpc.WebSocketMessageWriter(socket);
const socketConnection = server.createConnection(reader, writer, () => socket.dispose())
const serverConnection = server.createServerProcess('Example', 'node', ['example.js']);
server.forward(socketConnection, serverConnection, message => {
    if (rpc.isNotificationMessage(message)) {
        if (message.method === 'testNotification') {
            // handle the test notification
        }
    }
    return message;
});
```

But the method `isNotificationMessage()` doesn't exist in `rpc`. Am I missing something?

kaisalmen:
Hi @arnoson thank you for bringing this up. We forgot to update the README content. The example were updated in the meantime. Take a look at the [server example](https://github.com/TypeFox/monaco-languageclient/blob/main/packages/examples/server/) for now.

I will update the README content and will also add a link to the example where both `monaco-languageclient` and `vscode-ws-jsonrpc` are used.

arnoson:
Awesome, thanks for the quick reply!

kaisalmen:
It is now fixed on main branch, see #446

PR #446, opened 2023-01-06
Align vscode-ws-jsonrpc README and server example

kaisalmen:
This PR was triggered by #445 and corrects the example in the README.

I have polished the server example along and discovered something that we should discuss in another issue (#447).

### DIFF
--- a/packages/examples/server/src/server.ts
+++ b/packages/examples/server/src/server.ts
@@ -3,17 +3,13 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  * ------------------------------------------------------------------------------------------ */
 import { WebSocketServer } from 'ws';
-import * as http from 'http';
-import { fileURLToPath, URL } from 'url';
-import * as net from 'net';
+import { IncomingMessage } from 'http';
+import { URL } from 'url';
+import { Socket } from 'net';
 import express from 'express';
-import * as rpc from 'vscode-ws-jsonrpc';
+import { IWebSocket } from 'vscode-ws-jsonrpc';
 import { launch } from './json-server-launcher.js';
-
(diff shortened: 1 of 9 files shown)


packages/examples/server/src/server.ts
```diff
@@ -3,17 +3,12 @@
...
 import express from 'express';
-import * as rpc from 'vscode-ws-jsonrpc';
-import { launch } from './json-server-launcher.js';
-
-// solve: __dirname is not defined in ES module scope
-import path from 'path';
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
+import { IWebSocket } from 'vscode-ws-jsonrpc';
+import { getLocalDirectory, launch } from './json-server-launcher.js';
```
CGNonofr, line 11: I'm not sure `json-server-launcher` is the right place to put the `getLocalDirectory` method

kaisalmen: It is used there as well. Any good idea where to put it? New file and both reference that one?

CGNonofr: That's probably what I would have done, sbut I don't know how to name it :grin:

kaisalmen: Ok, I go for " iDontKnowHowToNameIt.ts". 😆 No, seriously, I do that and come up with a hopefully fitting name.

kaisalmen:
Done. Btw, are you generally ok with squash and rebase of PRs? I should not be a "must", but it helps to keep the git history graph a bit cleaner. It is not pressing here because of small project size. I just want to know what you think?

CGNonofr (approved):

PR #455, opened 2023-01-24
Gracefully handle JSON.parse errors

AlexTugarev:
This PR improves handling of parse errors for incoming message.

### DIFF
--- a/packages/vscode-ws-jsonrpc/src/socket/reader.ts
+++ b/packages/vscode-ws-jsonrpc/src/socket/reader.ts
@@ -60,8 +60,16 @@ export class WebSocketMessageReader extends AbstractMessageReader implements Mes
         if (this.state === 'initial') {
             this.events.splice(0, 0, { message });
         } else if (this.state === 'listening') {
-            const data = JSON.parse(message);
-            this.callback!(data);
+            try {
+                const data = JSON.parse(message);
+                this.callback!(data);
+            } catch (err) {
+                const error: Error = {
+                    name: '' + 400,
+                    message: `Error during message parsing, reason = ${typeof err === 'object' ? (err as any).message : 'unknown'}`
+                };
+                this.fireError(error);
+            }
         }
     }
 


packages/vscode-ws-jsonrpc/src/socket/reader.ts
```diff
@@ -60,8 +60,16 @@ export class WebSocketMessageReader extends AbstractMessageReader implements Mes
...
-            const data = JSON.parse(message);
-            this.callback!(data);
+            try {
+                const data = JSON.parse(message);
+                this.callback!(data);
+            } catch (err) {
+                const error: Error = {
+                    name: '' + 400,
+                    message: `Error during message parsing, reason = ${typeof err === 'object' ? (err as any).message : 'unknown'}`
+                };
```
CGNonofr, line 70: I have a few remarks:
- What about using the Error constructor?
- What does that `name` property means? and why 400 as a string?
- `typeof err === 'object'` doesn't mean it's an error, what about using `err instanceof Error`? (the cast wouldn't be required then

CGNonofr (changes requested):
Thanks for your contribution!

Were you experiencing issues or is this theoretical?

Or does this change comes from somewhere particular?

packages/vscode-ws-jsonrpc/src/socket/reader.ts
AlexTugarev: @CGNonofr, thanks for the questions. 'Trying to answer as much as possible. 

Here we're creating an error response, which is sent over the wire back to the client. This error is not to be further handled in this library. 
Also, this is a repetition of a pattern you can find above in this file: https://github.com/TypeFox/monaco-languageclient/blob/2087bd3ebe040ec9c5e986d7146625db64c607f0/packages/vscode-ws-jsonrpc/src/socket/reader.ts#L25

> What does that name property means? and why 400 as a string?

Error code `400` to signal a bad request. Again, following the example to show consistency, which should later on allow for improvement of similar cases. 

> doesn't mean it's an error

There isn't anything. Just trying to read the message from something that looks like an error, even if it wasn't constructed as an error. In any case, including 'unknown' is the fallback, and just in case we get to read a message, that will do.

CGNonofr:
@kaisalmen I'll let you have a look and approve, LGTM

CGNonofr (approved):

kaisalmen:
I have no objections. Who is going to push the button?

kaisalmen:
@AlexTugarev I guess you require a new release?

AlexTugarev:
👋🏻 @kaisalmen! 
Indeed, it would be awesome if`vscode-ws-jsonrpc` would be updated soon.

 🦊 🍻

kaisalmen:
Here you go: https://www.npmjs.com/package/vscode-ws-jsonrpc/v/2.0.2

PR #485, opened 2023-04-13
Update to monaco-vscode-api 1.78.0

kaisalmen:
As the title says the main dependency `monaco-vscode-api` has been lifted to version `1.78.0`.
`MonacoServices` and corresponding code has been removed and it is replaced by `initServices` exported by this lib serving as initialization tool for all services made available by `monaco-vscode-api`.

All examples have been adjusted and the angular example has been moved to its own [repository](https://github.com/TypeFox/monaco-languageclient-ng-example).

**This is outdated:**
@CGNonofr this is still a draft
I made the first examples work, but I have to at least initialize the following services otherwise I get errors in the log (see client, langium or react examples). Did you expect this?
```
    await initializeMonacoService({
        ...getModelEditorServiceOverride(async (model, options, sideBySide) => {
               return undefined;
        }),
        ...getNotificationServiceOverride(),
        ...getTextmateServiceOverride(),
        ...getThemeServiceOverride()
    });
    await initializeVscodeExtensions();
```
I killed several things from the lib as I understand they are no longer needed, like `MonacoServices`, but they need to be replaced with a **configurable** list of above services. This should be handled/supported by the lib and not by the user, I guess.


CGNonofr:
> @CGNonofr this is still a draft I made the first examples work, but I have to at least initialize the following services otherwise I get errors in the log (see client, langium or react examples). Did you expect this?
> 
> ```
>     await initializeMonacoService({
>         ...getModelEditorServiceOverride(async (model, options, sideBySide) => {
>                return undefined;
>         }),
>         ...getNotificationServiceOverride(),
>         ...getTextmateServiceOverride(),
>         ...getThemeServiceOverride()
>     });
>     await initializeVscodeExtensions();
> ```

Not really, but it's very hard to make all combinations work. There's no reason for getModelEditorServiceOverride and getNotificationServiceOverride to be mandatory indeed (it seems the first one is caused by a missing polyfill and the second one by cyclic dependency in the default impl)

> This should be handled/supported by the lib and not by the user, I guess.

What do you mean?

CGNonofr:
I've added the fixes (missing polyfill and cyclic dependency) and published a [1.78.0-next.1](https://www.npmjs.com/package/@codingame/monaco-vscode-api/v/1.78.0-next.1), can you try with it?

kaisalmen:
This error appears now in the webpack build:
image

CGNonofr:
> This error appears now in the webpack build: image

Did you change something else? because that issue is not new

kaisalmen:
> because that issue is not new

Do you have context? I am not aware of this issue. The verify vite build is fine. Only webpack has this problem.

CGNonofr:
The method is treeshaken from monaco and there is no way to polyfill it. I've open a PR to make it so it's not remove but they refuse to merge it. monaco-vscode-api exposes à webpack loader to polyfill it on the fly. It's used by the sniper service override. I think it fails now because you import it

kaisalmen:
It works, thank you!

kaisalmen:
The `initServices` is now working, but you are currently able to create erroneous service configs. The one below is the minimum stable one. Only the four services at the end which are set to false, can be activated or not without breaking anything. If I disable one of the other, there are either init errors or errors in the console during runtime.

```typescript
await initServices({
	enableNotificationService: true,
	enableThemeService: true,
	enableTextmateService: true,
	enableConfigurationService: true,
	configurationServiceConfig: {
		defaultWorkspaceUri: monaco.Uri.file('/')
	},
	enableKeybindingsService: true,
	enableDebugService: true,
	enableAudioCueService: true,
	enableDialogService: false,
	enableModelEditorService: false,
	modelEditorServiceConfig: {
		useDefaultFunction: true
	},
	enableLanguagesService: false,
	enablePreferencesService: false
}).then(() => createEditor());
```

kaisalmen:
@CGNonofr You can play around with it in one of the example. They are all working again.
We should find a minimum allowed services and always init those and only expose the optional services in that function above, I think.

CGNonofr:
Also it's not ideal for tree-shaking.

One solution would be to dynamically import them (but It would still generate a bundle even for user not using it)

kaisalmen:
> I'm not sure about the language service, don't you need it for langium? I have an error if I disable it

yes, because I register the statemachine language. It is required for the langium example, but not in the simple client example.

> The main issue here is that you always import all service overrides, but you use them or not based on a boolean. But importing the module already trigger some code. You can't just import a service-override without using the service.

Ok, that makes sense. 

> One solution would be to dynamically import them (but It would still generate a bundle even for user not using it)

Dynamic import works nicely. This way the lib users don't need to import all the things. And, there is no need to use the function. It's optional. You can still do it the other way. I added the top-level script `vite:build` that builds the example where you can see the dynamically loaded js files are seperate.

This could even be moved to monaco-vscode-api or stay here. WDYT about this solution?

Btw, the client example shows errors when nothing is configured.
image

CGNonofr:
> This could even be moved to monaco-vscode-api or stay here. WDYT about this solution?

I don't know, I think I prefer letting the user do what they want in monaco-vscode-api (for example provide their custom implementation for services like notifications or dialogs)

And this lib is there to simplify the usage

> Btw, the client example shows errors when nothing is configured. image

Yeah, it seems now the files service override is mandatory when using the output channel api

packages/client/src/monaco-vscode-api-services.ts
```diff
@@ -40,50 +32,112 @@ export type InitializeServiceConfig = {
...
+
+        if (config.enableDebugService === true) {
+            const { default: getDebugServiceOverride } = await import('vscode/service-override/debug');
+            allServices.debugService = getDebugServiceOverride();
+        }
+
+        if (config.enablePreferencesService === true) {
+            const { default: getPreferencesServiceOverride } = await import('vscode/service-override/preferences');
+            allServices.preferencesService = getPreferencesServiceOverride();
+        }
```
CGNonofr, line 140: It would probably be a good idea to load them all at once with a Promise.all()

also it would be simpler to just return a single `ServiceOverride`

kaisalmen: Yes, true. I will look into that.

kaisalmen:
> I don't know, I think I prefer letting the user do what they want in monaco-vscode-api (for example provide their custom implementation for services like notifications or dialogs)

Makes sense, I was kind of expecting this answer 🙂 

> And this lib is there to simplify the usage

Yes, but I could add another optional parameter that allows passing further services / own implementations.

> Yeah, it seems now the files service override is mandatory when using the output channel api

Ah, forgot about this new one.

kaisalmen:
@CGNonofr I have implement it properly now (Promise.all) and allow user services additionally (see langium example).

I did not manage to define a generic Promise of imports, that's why it is a bit hacky there, but it works nicely. If you have a better idea, please let me know. Verbose console logging is still included.

Browser example has and error which I did not yet fix.

CGNonofr:
Note that except `CallHierarchyFeature`, `TypeHierarchyFeature` and `NotebookDocumentSyncFeature`, all features can now be used I think (optionally)

And the first 2 may be supported soon as well

kaisalmen:
@CGNonofr I have moved the angular-client to its own repo now:
https://github.com/TypeFox/monaco-languageclient-ng-example

It polluted the dependencies and anyhow didn't integrate well into the vite workflow.

CGNonofr:
I agree, one demo is enough here to demonstrate the feature, other specific demos can be elsewhere

kaisalmen:
I have release a next version:
https://www.npmjs.com/package/monaco-languageclient/v/5.1.0-next.0

kaisalmen:
@CGNonofr do you have any idea why this warning is shown when I load the configuration service dynamically?
image

When I do it like this it works without warning:
```
await initServices({
    userServices: {
        ...getConfigurationServiceOverride(monaco.Uri.file('/')),
    }
})
```
With the debugger in both cases I see that the `MemoryFileService` constructor is reached, but only when I use the dynamic import the warning is shown. I feel like I am chasing ghosts. Do you have any idea what's going wrong here? Thanks.

CGNonofr:
> @CGNonofr do you have any idea why this warning is shown when I load the configuration service dynamically? image
> 
> When I do it like this it works without warning:
> 
> ```
> await initServices({
>     userServices: {
>         ...getConfigurationServiceOverride(monaco.Uri.file('/')),
>     }
> })
> ```
> 
> With the debugger in both cases I see that the `MemoryFileService` constructor is reached, but only when I use the dynamic import the warning is shown. I feel like I am chasing ghosts. Do you have any idea what's going wrong here? Thanks.

Yep the error is misleading, it just tries to stats the workspace folder to check if it's a folder or a file and is returned a Not found error. It has no consequences, I have the same issue and I'll try to find a solution to remove this log warn

kaisalmen:
> I have the same issue and I'll try to find a solution to remove this log warn

Good to know. I put the console logging in monaco-vscode-api-services behind a flag now. It is use in Langium example only. Did you already had the chance to look at the code and test it?

Question: With the Web Worker Language Server capability does it make sense to keep the old browser example? WDYT?

kaisalmen:
@CGNonofr client code looks good now. I have polished the code and removed the need for `typeof import` hacks

kaisalmen:
@CGNonofr are you busy currently or will you find time this week to look at this? Thank you.

packages/client/src/monaco-vscode-api-services.ts
```diff
@@ -0,0 +1,162 @@
...
+                console.log('initializeMonacoService completed successfully');
+            }
+        })
+        .then(async () => await initializeVscodeExtensions())
+        .then(() => {
+            if (config?.debugLogging === true) {
+                console.log('initializeVscodeExtensions completed successfully');
+            }
+        })
+        .catch((e: Error) => { throw e; });
```
CGNonofr, line 47: What about using async/await?

packages/examples/main/src/langium/main.ts
```diff
@@ -149,6 +121,29 @@ const run = async () => {
...
+    },
+    enableKeybindingsService: false,
+    enableLanguagesService: true,
+    enableAudioCueService: true,
+    enableDebugService: true,
+    enableDialogService: true,
+    enableNotificationService: true,
+    enablePreferencesService: true,
+    userServices: {
+        ...getKeybindingsServiceOverride()
```
CGNonofr, line 143: Why enabling it by hands?

packages/examples/main/src/react/app.tsx
```diff
@@ -92,31 +81,41 @@ export const ReactMonacoEditor: React.FC<EditorProps> = ({
...
+            const monacoVscodeApiInit = async () => {
+                await initServices({
+                    enableThemeService: true
+                }).then(() => createEditor());
+            };
+            if (init) {
+                monacoVscodeApiInit();
+            } else {
+                createEditor();
+            }
```
CGNonofr, line 115: Where is init set to false?

Also the syntax pretty confusing

what about:
```typescript
const initAndCreateEditor = async () => {
  if (init) {
    await initServices({
      enableThemeService: true
    })
    init = false
  }
  createEditor();
}
```

or even better:

```typescript
const vscodeApiInitPromise = initServices({
  enableThemeService: true
})

[...]

  await vscodeApiInitPromise
  createEditor();
```

packages/examples/main/src/react/app.tsx
```diff
@@ -92,31 +81,41 @@ export const ReactMonacoEditor: React.FC<EditorProps> = ({
...
 
             return () => {
-                editorRef.current!.dispose();
+                editorRef.current?.dispose();
```
CGNonofr, line 118: nit: It's not a good practice to use a ref inside the effect dispose: better store the ref value at the effect start and use it here

packages/examples/main/src/langium/main.ts
kaisalmen: Just to show it is possible to doing it both ways. A comment would not hurt here.

packages/client/src/monaco-vscode-api-services.ts
kaisalmen: You mean just write await statements and not intercept the Error here an let the next level deal with it catch it?
```Typescript
export const initServices = async (config?: InitializeServiceConfig) => {
    await importAllServices(config);
    if (config?.debugLogging === true) {
        console.log('initializeMonacoService completed successfully');
    }
    await initializeVscodeExtensions();
    if (config?.debugLogging === true) {
        console.log('initializeVscodeExtensions completed successfully');
    }
};
```

CGNonofr: My comment doesn't concern the error management, it is about not using `.then` or `.catch`

btw, `.catch((e: Error) => { throw e; });` seems to be pointless, isn't it?

packages/examples/main/src/react/app.tsx
kaisalmen: Thanks, I am not very deep into react.

packages/client/src/monaco-vscode-api-services.ts
kaisalmen: > My comment doesn't concern the error management, it is about not using .then or .catch

Yes, I got this. I wanted to chain the catch, but using a catch is fine, but here it does not even make sense, because what was there is pointless.

packages/examples/main/src/react/app.tsx
kaisalmen: What about this:
```typescript
if (init) {
	(async () => {
		await initServices({
			enableThemeService: true
		});
		createEditor();
		init = false;
	})();
} else {
	createEditor();
}
```

packages/examples/main/src/react/app.tsx
kaisalmen: Implemented as suggested

packages/examples/main/src/langium/main.ts
kaisalmen: Added a comment, see commit below ⬇️

packages/client/src/monaco-vscode-api-services.ts
kaisalmen: Changed as suggested. Remove superfluous catch/re-throw

packages/examples/main/src/react/app.tsx
kaisalmen: Implemented as pointed out ⬆️, see commit below ⬇️

kaisalmen:
@CGNonofr async code is in better shape now. I added `vscodeApiInitialised` into global `MonacoEnvironment` as I need to know if the services were initialized when using multiple monaco-editors on a page.

kaisalmen:
@CGNonofr do you think this is good now as well?
On Windows the 'vscode/webpack-loader' does not work and it seems to be related to the path separators. I will open a PR, I guess:
image

kaisalmen:
> I never work on windows

I like both and work on both. 😎

packages/examples/main/src/browser/main.ts
```diff
@@ -11,145 +11,153 @@ import { buildWorkerDefinition } from 'monaco-editor-workers';
...
+        model: modelRef.object.textEditorModel,
+        glyphMargin: true,
+        lightbulb: {
+            enabled: true
+        },
+        automaticLayout: true
+    });
 
-const vscodeDocument = vscode.workspace.textDocuments[0];
+    const mainVscodeDocument = await vscode.workspace.openTextDocument({
```
kaisalmen, line 51: @CGNonofr this now longer works
```javascript
const vscodeDocument = vscode.workspace.textDocuments[0];
```
But my replacement throws an error:
image
Do you have an idea how to fix this? Will https://github.com/CodinGame/monaco-vscode-api/pull/109 be of help here, too?

CGNonofr: Can you please add this comment on the other side (on the PR) to not forget about it?

CGNonofr: Btw i'm not sure to understand why you want to open an untitled text document

kaisalmen: Like this? https://github.com/CodinGame/monaco-vscode-api/pull/109#issuecomment-1531551700

kaisalmen: > Btw i'm not sure to understand why you want to open an untitled text document

It seems in the past a default document was available. Is there a way to create a `vscode.TextDocument` from the already existing textEditorModel or is it actually wrong to use the workspace here.

The real question is if it is worth to keep the browser example at all if we have web worker based language servers now, but still I want to understand/fix this problem here?

CGNonofr: vscode.workspace.textDocuments returns a vscode.TextDocument for each open model, I don't know what what that "default" document that was there before

maybe you're calling it too early, the synchronization between the extHost and the services was made asynchronous, maybe you can wait on `vscode.workspace.onDidOpenTextDocument`

kaisalmen: > maybe you're calling it too early, the synchronization between the extHost and the services was made asynchronous, maybe you can wait on vscode.workspace.onDidOpenTextDocument

This could explain it. I will give it a try later.

kaisalmen: Fixed it, see below ⬇️

kaisalmen:
@CGNonofr after upgrading to version 1.78.5-next.0 both verification builds for webpack and vite break. I removed the webpack-loader as well.

kaisalmen:
@CGNonofr once https://github.com/CodinGame/monaco-vscode-api/pull/109 is merged, I would like to merge this PR in as well, but not yet immediately release a new version of the lib. This is already a huge change and taking a look on what has changed the next version will be a major one.

CGNonofr:
I guess some languageclient features can be enabled back, like WillSaveFeature, WillSaveWaitUntilFeature and DidSaveTextDocumentFeature

With https://github.com/CodinGame/monaco-vscode-api/pull/110, it will probably make sense to enable almost all features except notebook and progress ones

kaisalmen:
> I guess some languageclient features can be enabled back

Good point. I forgot about that.

kaisalmen:
@CGNonofr features are now registered now ⬆️ I left comments on what will become available soon.

Btw, I have removed redundant code from the various client examples (now in common.ts). This is ready to go. I will now release `6.0.0-next.0`

CGNonofr (approved):
LGTM

kaisalmen:
@CGNonofr can you check the first three section of the README?
I have added "May 2023" to
https://github.com/TypeFox/monaco-languageclient#latest-important-project-changes

Added new chapter "Using monaco-languageclient":
https://github.com/TypeFox/monaco-languageclient#using-monaco-languageclient

and moved "Monaco-editor compatibility table" up:
https://github.com/TypeFox/monaco-languageclient#monaco-editor-compatibility-table

If that is fine I will perform the release. Thank you!

CGNonofr:
Very fine! I wish the documentation of monaco-vscode-api was as good as yours

kaisalmen:
Released:
https://www.npmjs.com/package/monaco-languageclient/v/6.0.0

> Very fine! I wish the documentation of monaco-vscode-api was as good as yours

It evolved over time! 😎 I think the documentation of monaco-vscode-api is already quite good. We need to add unit test here and there 😆 🙈

BusinessDuck:
Hi guys, seems like patch didn't work. I don't know why, but script says "patch didn't applied". :(

CGNonofr:
It's probably because you have a mismatch with the monaco version

kaisalmen:
@BusinessDuck  Do you use monaco-editor 0.37.1?
@CGNonofr because of the patch the peerDependency to monaco-editor in monaco-vscode-api must be specific, but it is still `~0.37.1`. This is not a problem as there won't be a 0.37.2, but we should correct this for next major release 1.79.x (June)

BusinessDuck:
>@BusinessDuck Do you use monaco-editor 0.37.1?

Yes, i am working with editor-core and alias for it

package.json listing:

```
"monaco-editor": "npm:monaco-editor-core@0.37.1",
"monaco-editor-core": "0.37.1",
```

BusinessDuck:
How i can enable verbose for monaco-treemending script?

Now error look lije this (not enough details to get reason):
```
Downloading registry.npmjs.org/monaco-editor-core/0.37.1: 10.3 MB/10.3 MB, done
. postinstall$ monaco-treemending
│ Error: Unable to apply patch on ....
```

CGNonofr:
monaco-editor already include monaco-editor-core, it's probably a bad idea to have both

there is a patch file at the root of monaco-vscode-api, you can try applying it by hands with `patch -p1 < file.patch` from monaco-editor/esm directory

BusinessDuck:
>monaco-editor already includes monaco-editor-core, it's probably a bad idea to have both

That because of some packages may contains imports from `monaco-editor` this is wrong now, because `monaco-editor-core` is low level package (proprietary) all dev tools like vscode-api and other languageclient helpers shouldn't use imports from `monaco-editor` microsoft force using editor core for any cases when we dont need access to out of the box languages and other monaco-editor's stuff

kaisalmen:
Why can't you do like this. We updated all examples to only load the content of monaco-editor-core. It is really equivalent:
```javascript
import * as monaco from 'monaco-editor/esm/vs/editor/edcore.main.js';
```

BusinessDuck:
>there is a patch file at the root of monaco-vscode-api, you can try applying it by hands with patch -p1 < file.patch from monaco-editor/esm directory

No output, command looks stalled. Platform Macos

Also i have remove `monaco-editor-core` and convert all usages to `monaco-editor`, didn't help to me :(

BusinessDuck:
all relates package list:

```
"@types/vscode": "1.78.0",
"vscode": "npm:@codingame/monaco-vscode-api@1.78.5",
"vscode-jsonrpc": "8.1.0",
"vscode-languageclient": "~8.1.0",
"monaco-editor": "0.37.1",
"monaco-languageclient": "6.0.0",
    ```

**Comment by BusinessDuck** [comment]
image
I attach debugger for patching process, and see error here

**Comment by kaisalmen** [comment]
I don't use pnpm, so I have no experience solving problems there.
Did you try to purge the node_modules folder? Is it a workspace project or just one project. Is that publicly available?
I tested the patch yesterday [here](https://github.com/TypeFox/monaco-components/tree/vscode-config) and just updated to 6.0.0. Pipeline works. Works both on Linux and Windows, but using npm.

**Comment by BusinessDuck** [comment]
>Did you try to purge the node_modules folder? 
I did, sure :) That first what i didi actually :)

**Comment by kaisalmen** [comment]
Have you actually checked out this repo and tried to build it completely?

**Comment by BusinessDuck** [comment]
image
Failed index details also.

>Have you actually checked out this repo and tried to build it completely?

one moment

**Comment by kaisalmen** [comment]
It applies the treemending-patch as well that's why I am asking.
Doesn't pnpm cache all modules centrally, maybe that's causing the patching issues.

**Comment by BusinessDuck** [comment]
In current repo by using `npm i` in pwd `monaco-languageclient/packages/client` it works great. But when i try to run `npm i` second time then postinstall falls with same error.


For the first time using it was success `Monaco-editor was tree-mended`

**Comment by kaisalmen** [comment]
Ok, I can confirm that when triggering it manually. When I execute `npm i` multiple times this error is suppressed. How do you invoke the treemending script?

**Comment by kaisalmen** [comment]
Do you use it as postinstall like suggested?
```json
{
  "scripts": {
    "postinstall": "monaco-treemending",
  }
}
```

BusinessDuck:
OMG! It's `npm` itself, i have update to latest and errors i gone....  My npm version is before update 9.0.1 and after 9.6.6

CGNonofr:
Feel free to implement something, I'm not sure how to do it

kaisalmen:
> Feel free to implement something, I'm not sure how to do it

I will think about it, do some research. This will not be the first time someone has this kind of problem.

kaisalmen:
@BusinessDuck within this repo: https://github.com/TypeFox/monaco-languageclient/blob/main/packages/examples/main/src/langium/main.ts shows best what is possible and how to use `initServices`. Or check the demo in https://github.com/CodinGame/monaco-vscode-api/tree/main/demo

PR #501, opened 2023-06-06
Adapt to monaco-vscode-api-1.79.1

kaisalmen:
@CGNonofr I have a problem even if I only include files and model. Any idea on this?
image

I changed `initServices` that users can only include editor or views as you outlined in the monaco-vscode-api README

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -23,35 +23,7 @@ export default defineConfig(() => {
             port: 8080,
             origin: 'http://localhost:8080'
         },
-        assetsInclude: ['**/*.wasm'],
-        optimizeDeps: [
-
-            'monaco-editor',
-            'vscode',
-            'vscode/extensions',
-            'vscode/services',
-            'vscode/monaco',
-            'vscode/service-override/model',
-            'vscode/service-override/editor',
-            'vscode/service-override/notifications',
-            'vscode/service-override/dialogs',
-            'vscode/service-override/configuration',
-            'vscode/service-override/keybindings',
-            'vscode/service-override/textmate',
-            'vscode/service-override/theme',
-            'vscode/service-override/languages',
-            'vscode/service-override/audioCue',
-            'vscode/service-override/views',
-            'vscode/service-override/quickaccess',
-            'vscode/service-override/debug',
-            'vscode/service-override/preferences',
-            'vscode/service-override/snippets',
-            'vscode/service-override/files',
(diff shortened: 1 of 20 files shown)


CGNonofr:
monaco-vscode-api includes the reference search from vscode
so `monaco-editor/esm/vs/editor/standalone/browser/referenceSearch/standaloneReferenceSearch` shouldn't be included, nr directly `monaco-editor` obviously

kaisalmen:
This means importing:
```javascipt
import 'monaco-editor/esm/vs/editor/edcore.main.js'
```
is now forbidden

CGNonofr:
It seems so!

btw, it was already a bad idea, as most of them are not functionnal, like `toggleHighContrast` and `inspectTokens`.

At least now we can't make the mistake or importing it (:

kaisalmen:
This is ready. I just released: https://www.npmjs.com/package/monaco-languageclient/v/6.1.0-next.1

CGNonofr:
I don't know it it's a good idea to implicitly includes other services (keybindings from quickaccess)

CGNonofr:
> > I don't know it it's a good idea to implicitly includes other services (keybindings from quickaccess)
> 
> At least better than not working. Do you prefer throwing errors suggesting what to do?

I think it's preferable

kaisalmen:
I am 50/50 on that, but with more services (I have seen the next branch with more services 😱 😆 ) and growing complexity it might be a better idea.

kaisalmen:
@CGNonofr now errors are thrown. No dependent services are configured automatically. I kept the files service being always loaded. Any further objections?

CGNonofr:
> now errors are thrown

you mean `no`?

vite.config.ts
```diff
@@ -23,7 +23,35 @@ export default defineConfig(() => {
...
+            'vscode/service-override/views',
+            'vscode/service-override/quickaccess',
+            'vscode/service-override/debug',
+            'vscode/service-override/preferences',
+            'vscode/service-override/snippets',
+            'vscode/service-override/files',
+            'vscode/default-extensions/theme-defaults',
+            'vscode/default-extensions/javascript',
+            'vscode/default-extensions/typescript',
+            'vscode/default-extensions/json'
```
CGNonofr, line 53: Do you need to do that? I was thinking it's required only in the vscode-api demo because it's a local dependency

CGNonofr (approved):

PR #503, opened 2023-06-13
Updated to monaco-vscode-api 1.79.2 / monaco-editor 0.39.0

kaisalmen:
Still WIP.
Even without adding new services, I see this error now:
image

Is this already fixed by this: https://github.com/CodinGame/monaco-vscode-api/pull/128 ?


kaisalmen:
@CGNonofr I guess it makes sense to release a next version with every big monaco-vscode-api PR and test it here beforehand as well.

CGNonofr:
> Is this already fixed by this: https://github.com/CodinGame/monaco-vscode-api/pull/128 ?

Probably not! I'll have a look

CGNonofr:
Ok it's fixed in https://github.com/CodinGame/monaco-vscode-api/pull/128/commits/824f51bdb3f19997272ae5dd72dc20bc1a0ef5d8

kaisalmen:
First problem is gone. If I now just add the markers service. Does it depend on something else? Here I added everything, but the new terminal service. Does this even make sense without using the view service?
image

kaisalmen:
@CGNonofr FYI I just copied over your TerminalBackend from the demo and then I get two errors (from markers and terminal):
image

CGNonofr:
> First problem is gone. This error shows up if I add the markers service. Does it depend on something else? n't configure the markers and terminal service there is no error.

I guess it won't work without the view services, but the markers service override just adds the `Problems` view so it has no point anyway

> Here I added everything, but the new terminal service. Does this even make sense without using the view service?

No. The marker service override doesn't add any service for the moment, it just register the marker view

> FYI I just copied over your TerminalBackend from the demo and then I get two errors (from markers and terminal):

THAT is weird, I'll have a look

kaisalmen:
I wanted to know if that is your expectation, too. I think that this is the correct behaviour.

CGNonofr:
Maybe we can improve the doc but importing markers without the views make no sense

kaisalmen:
> Maybe we can improve the doc but importing markers without the views make no sense

Yes, docs should be updated. I added this code: https://github.com/TypeFox/monaco-languageclient/blob/mva-1.79.2/packages/client/src/monaco-vscode-api-services.ts#L171-L191 to flag produce errors on bad configuration.

kaisalmen:
@CGNonofr `1.79.3-next.4` works fine. 👍
I can confirm that when I add a panel and use markers with views service there are no more errors.

kaisalmen:
@CGNonofr are you ok with changes. I will update the version to 6.2.0 after merge.

kaisalmen:
Oh it was still a draft. Version change is in here already.

kaisalmen:
@CGNonofr you raised a thumb, but you have to approve. Thank you

kaisalmen:
I went ahead. Want to publish the release. 😊

kaisalmen:
Sorry, I overlooked that 🙂

PR #520, opened 2023-08-09
Update to monaco-vscode-api 1.81.0 and monaco-editor 0.41.0

kaisalmen:
This is still WIP.


CGNonofr:
One day we'll be able to update to a version before the next one is released :grimacing:

kaisalmen:
> One day we'll be able to update to a version before the next one is released 😬

Next month! Promised or we try the month after. 😆

kaisalmen:
@CGNonofr as soon as you release final `monaco-vscode-api@1.81.0` I will update this again and then you could approve. I tested the preview release (https://www.npmjs.com/package/monaco-languageclient/v/6.4.0-next.0) an it is looking good.

packages/examples/main/src/browser/main.ts
```diff
@@ -121,7 +121,7 @@ const createEditor = async () => {
         validate();
     });
 
-    workspace.onDidOpenTextDocument((_event) => {
+    workspace.onDidChangeTextDocument((_event) => {
```
CGNonofr, line 124: was it done on purpose?

packages/examples/main/src/langium/statemachineClient.ts
```diff
@@ -146,7 +145,6 @@ try {
         debugLogging: true,
         logLevel: LogLevel.Info
     });
-    // renderPanelPart(document.querySelector<HTMLDivElement>('#panel')!);
```
CGNonofr, line 149: Do you want to keep the comments?

packages/examples/main/src/browser/main.ts
kaisalmen: This was unstable. Sometime it worked. Sometime it didn't. I already forgot I changed this last night. Should we investigate?

packages/examples/main/src/langium/statemachineClient.ts
kaisalmen: No, I removed them for now. Thinking about actually adding another example that could make use of views (e.g. #521)

packages/examples/main/src/browser/main.ts
CGNonofr: isn't it a race condition? you should probably search for your file inside workspace.textDocuments

kaisalmen: ok, there seems to be a problem `onDidOpenTextDocument` is really never called. I added log statements locally and I don't see it. Will try in other examples. Maybe it broke earlier. The example still work, but you see errors in the log.

kaisalmen: My fault, it is the order. The callback `onDidOpenTextDocument`. is added to late. I will fix this

kaisalmen: Done ⬇️

kaisalmen:
@CGNonofr Ready

packages/examples/main/src/browser/main.ts
```diff
@@ -83,6 +89,7 @@ const createEditor = async () => {
     });
 
     const validate = () => {
+        console.log('validate');
```
CGNonofr, line 92: Was it forgotten?

kaisalmen: Thanks, fixed ⬇️

CGNonofr (approved):

PR #523, opened 2023-08-16
Update to monaco-vscode-api 1.81.1

kaisalmen:
Looks good. If you release the final version of `monaco-vscode-api@1.81.1`, I will update this one here:
- Service configuration was simplified due to the changes in `monaco-vscode-api`
- Removed `enableQuickaccessService` as it is always loaded now as default
- Aligned naming of properties in `InitializeServiceConfig`


packages/client/src/monaco-vscode-api-services.ts
```diff
@@ -35,15 +36,17 @@ export type InitializeServiceConfig = {
...
     enableOutputService?: boolean;
-    configureTerminalServiceConfig?: {
+    configureTerminalService?: {
         backendImpl: SimpleTerminalBackend | ITerminalBackend
     }
     enableSearchService?: boolean;
     enableMarkersService?: boolean;
     enableAccessibilityService?: boolean;
     enableLanguageDetectionWorkerService?: boolean;
+    configureStorateService?: {
```
CGNonofr, line 47: Stora**t**e

kaisalmen: Thanks, corrected ⬇️

kaisalmen:
@CGNonofr. I was wondering why passing the messageTransport is overly complicated in `MonacoLanguaclient`. Was thinking if we should extend it from `vscode-languageclient/lib/browser/main.js` and then just override the createMessageTransports with a direct `MessageTransports` passed in the constructor, but that would bump the version 6.5.0. WDYT?

CGNonofr:
> @CGNonofr. I was wondering why passing the messageTransport is overly complicated in `MonacoLanguaclient`. Was thinking if we should extend it from `vscode-languageclient/lib/browser/main.js` and then just override the createMessageTransports with a direct `MessageTransports` passed in the constructor, but that would bump the version 6.5.0. WDYT?

I'm not sure what you mean, wouln't it prevent the user from using the vscode-language client reconnect feature? (creating a new connection each time the provider is called) we're using this feature

kaisalmen:
@CGNonofr I mean it like this: https://github.com/TypeFox/monaco-languageclient/tree/alt-constructor
In the end the `createMessageTransports` is invoked. What am I missing? But actually let's discuss it separately. That's why I put it on on a new branch.

Any objections to merge this?

CGNonofr:
> Any objections to merge this?

Nop!

CGNonofr (approved):

Issue #521, opened 2023-08-10
Add pyright languageclient / languageserver example


kaisalmen:
People asked about more complex languageclient examples multiple time and I think porting https://github.com/microsoft/pyright/tree/main/packages/vscode-pyright is the best target. We should also be able to add the node based LSP fairly easily here.
This is a far better target (maintenance-wise and overall work-wise) than then a Java languageclient.

kaisalmen:
@CGNonofr I started this. Connecting, but nothing more, yet.

kaisalmen:
Code completion is now working:
![image](https://github.com/TypeFox/monaco-languageclient/assets/21079138/be31adc6-2856-4294-8cce-1c8e68e00f1b)

snoop244:
Cheering you on @kaisalmen ! I completely agree about the need for a more substantial example. I've been checking in on the great TypeFox "glue" between Monaco and language servers over the last few years but I'm just not smart enought to make the leap from a JSON language server to a Python or Go language server.  

I'll monitor progress and test if that helps.

kaisalmen:
I opened a draft PR. It requires some polishing and further documentation, but can already be tested.

kaisalmen:
#526 has been merged.

PR #526, opened 2023-08-21
Python Language Client and Language Server example

kaisalmen:
Implements #521 :
- Adds the pyright language server (LS) that is started by an express app which handles the web socket connection to the client and forwards request to the python language server (similar to the JSON server example). The pyright LS runs as separate node process
- Adds python language client that is initialized as vscode extension
- Removed JSON server external launcher as this is now available with python LS
- Moved JSON client and server to sub-directory
- Update README and CHANGELOG
- Aligned code in langium examples to python client

Additional configuration from the pyright vscode extension could be included in the future. I wanted to get it going for now.


kaisalmen:
@CGNonofr I changed the configuration as you suggested. This is now ready for review.

packages/examples/main/src/python/client.ts
```diff
@@ -0,0 +1,153 @@
...
+        languageClient = createLanguageClient({
+            reader,
+            writer
+        });
+        await languageClient.start();
+        reader.onClose(() => languageClient.stop());
+
+        vscode.commands.registerCommand('pyright.restartserver', (...args: unknown[]) => {
+            languageClient.sendRequest('workspace/executeCommand', { command: 'pyright.restartserver', arguments: args });
+        });
```
CGNonofr, line 42: What if the socket is recreated? is the command overriden?

Maybe storing the last languageClient in a variable and registering the command at the root is more common

CGNonofr (commented):
Works great, good work!

It's just a shame that the ctrl+click on the `print` function doesn't work because it can access the filesystem. it would probably be great to register an underlay filesystem that have access to the node_modules, either by hands or using the HTMLFileSystemProvider

packages/examples/main/src/python/client.ts
kaisalmen: True, I moved the registration to the run function and extracted the function. This way it always uses the current languageclient.

kaisalmen:
> it would probably be great to register an underlay filesystem that have access to the node_modules, either by hands or using the HTMLFileSystemProvider

@CGNonofr Don't know if this would work in a distributed context (web client + node server not on the same platform). Maybe this can be simplified by configuration of the pyright server (possible module resolution path), but I don't know currently if that will work. Then we could ensure client and server have the same URIs for resolution.

Btw, I improved console logging in the server. Makes me wonder if I should revert the change to `vscode-ws-jsonrpc`. WDYT?

CGNonofr:
> > it would probably be great to register an underlay filesystem that have access to the node_modules, either by hands or using the HTMLFileSystemProvider
> 
> @CGNonofr Don't know if this would work in a distributed context (web client + node server not on the same platform).

Indeed it won't, we then need the server to expose an endpoint to retrieve any file content

> Maybe this can be simplified by configuration of the pyright server (possible module resolution path), but I don't know currently if that will work. Then we could ensure client and server have the same URIs for resolution.

Not exactly sure what you mean here

> Btw, I improved console logging in the server. Makes me wonder if I should revert the change to `vscode-ws-jsonrpc`. WDYT?

Why do you need the process for? you don't seem to use it anywhere

kaisalmen:
> Not exactly sure what you mean here

I meant if we can ensure both sides have the same content in the workspace / their file system. Currently the server provides the link location `.../node_modules/pyright/dist/typeshed-fallback/stdlib/builtins.pyi`

> Indeed it won't, we then need the server to expose an endpoint to retrieve any file content

Can you explain what you mean by endpoint here?

> Why do you need the process for? you don't seem to use it anywhere

I did log stdout before, but no longer do. I will revert that change

CGNonofr:
> I meant if we can ensure both sides have the same content in the workspace / their file system. Currently the server provides the link location .../node_modules/pyright/dist/typeshed-fallback/stdlib/builtins.pyi

What do you mean by a client having a workspace or a filesystem? the client is only able to access what you provide to it, either by writing a file in the memory filesystem or by registering an overlay filesystem

> Can you explain what you mean by endpoint here?

If you want the client to display files stored on the server, the server will need to expose a way for the client to retrieve those files

kaisalmen:
> What do you mean by a client having a workspace or a filesystem? the client is only able to access what you provide to it, either by writing a file in the memory filesystem or by registering an overlay filesystem

As this is not purely dynamic we could provide all file needed in the memory filesystem. It would be easier if all those extra files (like pyi files) are located at a configurable place on the server size, so we can mirror that on the client. Is that clearer?

> If you want the client to display files stored on the server, the server will need to expose a way for the client to retrieve those files

Ok, got it

kaisalmen:
@CGNonofr reverted the vscode-ws-jsonrpc changes. IMO, we should extend functionality in a second PR and merge this first. WDYT?

PR #532, opened 2023-08-29
Additional parameters for the URL creation

Stuey61296:
- Added _searchParams_ parameter to the createUrl function
- Extended the python example to send in an authorization parameter
- Added verifyClient function for authorization & client tracking to the language server


kaisalmen (approved):
LGTM. Thank you.

PR #549, opened 2023-10-11
Update to monaco-vscode-api 1.82.5

kaisalmen:
WIP: The server changes are not working as expected.

What the title says
In addition, I unified the language server usage. Less code to maintain.

I have the strange observation that the first code completion request from the JSON LS takes a long time, but afterwards works fine. I don't remember it used to be like this. 🤷‍♂️ Can you test this as well? Thanks.

Will update once https://github.com/CodinGame/monaco-vscode-api/pull/206 is merged and final release is available.


CGNonofr:
LGTM

CGNonofr:
> I have the strange observation that the first code completion request from the JSON LS takes a long time, but afterwards works fine. I don't remember it used to be like this. 🤷‍♂️ Can you test this as well? Thanks.

I have the same issue, the server doesn't answer to hover and completion before ~15s

kaisalmen:
> I have the same issue, the server doesn't answer to hover and completion requests before ~15s

This was my Don Quijote moment today. Fighting the windmill. I went back to 6.5.0 and had the same issues. What was definitely not the case two weeks ago.

I will merge this change now and iterate from there (upgrading to the funak

kaisalmen:
@CGNonofr Is `vscode-json-languageservice` phoning home or doing other internal BS? 💩 
It now works without problem 🧐 I updated to the `1.82.5-next.0` but that code is not involved with the language server. This stinks a bit.

kaisalmen:
Could it be linked to this? https://github.com/microsoft/vscode-json-languageservice/issues/212

CGNonofr:
It's probably related to some downloading indeed, maybe the schema server was slow

kaisalmen:
> It's probably related to some downloading indeed, maybe the schema server was slow

Very likely the case. Timing was very bad. Because I thought I broke something last night / this morning.
Good thing is: The server start-up is now unified.

PR #600, opened 2024-01-29
Merge monaco-editor-wrapper and monaco-editor-react into monaco-languageclient's npm workspace

kaisalmen:
This PR is now ready for review. A couple of open points should be fixed, but review can start now.

This repository now hosts the following npm packages:
- `monaco-languagclient`; Next release: https://www.npmjs.com/package/monaco-languageclient/v/8.0.0-next.5
- `vscode-ws-jsopnrpc`; Next release: https://www.npmjs.com/package/vscode-ws-jsonrpc/v/3.3.0-next.5
- `monaco-editor-wrapper`; Next release: https://www.npmjs.com/package/monaco-editor-wrapper/v/4.0.0-next.5
- `@typefox/monaco-editor-react`; Next release: https://www.npmjs.com/package/@typefox/monaco-editor-react/v/3.0.0-next.5
- `monaco-languagclient-examples`; Next release: https://www.npmjs.com/package/monaco-languageclient-examples/v/8.0.0-next.5

Important points: 
- The examples of this repository and the ones from https://github.com/TypeFox/monaco-components have been merged.
- Verification examples have been removed from the npm workspace and are now contained in the directory **verify**.
  - The Angular example has been integrated here. The build pipeline verifies they are properly building
  - They have been excluded from main workspace to prevent pollution of the main node_modules (e.g. angular)
- `monaco-editor-workers` is no longer needed and is replaced by `monaco-editor-wrapper/workerFactory` that integrates ideas from monaco-vscode-api demo and is fully customizable/over-writeable.
- Client examples now make use of  `monaco-editor-wrapper` or `@typefox/monaco-editor-react`. This made reduces the amount of boiler plate code.

Left over points required at least required to be implemented before this PR can be merged:

- [x] Browser example should also use the wrapper
- [x] Update all READMEs and linked documentation files
- [x] Check LICENSE files and statements
- [x] Update CHANGELOGs
- [x] Check if existing bugs should be fixed along or if they should be fixed with a bug-fix release
- [ ] The worker based extension host needs to be integrate to allow advanced TypeScript support where needed in the example (wrapperTs and wrapperAdvanced), but this can be deferred.

Will close #543

PR #602, opened 2024-02-13
Implement dispose method on WebSocketMessageReader.

ethan-vanderheijden:
Implementing `dispose()` according to the spec:
https://github.com/microsoft/vscode-languageserver-node/blob/21f4f0af6bf1623483c3e65f36e550bfdb6245a6/jsonrpc/src/common/messageReader.ts#L48

Although the `listen()` method returns a disposable callback, it seems like it's not used. Instead, the `dispose()` method is called on the reader (see [here](https://github.com/microsoft/vscode-languageserver-node/blob/21f4f0af6bf1623483c3e65f36e550bfdb6245a6/jsonrpc/src/common/connection.ts#L1553)).

Without implementing the dispose method, you can stop the language client, but calling `languageClient.start()` will not restart it properly.

### DIFF
--- a/packages/vscode-ws-jsonrpc/src/socket/reader.ts
+++ b/packages/vscode-ws-jsonrpc/src/socket/reader.ts
@@ -53,10 +53,18 @@ export class WebSocketMessageReader extends AbstractMessageReader implements Mes
         return {
             dispose: () => {
                 if (this.callback === callback) {
+                    this.state = 'initial';
                     this.callback = undefined;
                 }
             }
         };
+    }
+
+    override dispose() {
+        super.dispose();
+        this.state = 'initial';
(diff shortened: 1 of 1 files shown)


kaisalmen (approved):
That makes sense. Thank you.

ethan-vanderheijden:
Sorry, can you do a new release of `vscode-ws-jsonrpc`?

kaisalmen:
@ethan-vanderheijden yes, good point. I wanted to put out the next release when everything else is ready, but #600 is not yet done. A new version should be there end of the day.

kaisalmen:
@ethan-vanderheijden https://www.npmjs.com/package/vscode-ws-jsonrpc/v/3.2.0 is now available.

PR #629, opened 2024-04-09
Updated to monaco-vscode-api libs version 4

kaisalmen:
@CGNonofr it is all working with the exception of the new `getLocalizationServiceOverride`. Once I configure it, I see a stack of errors. The Statemachine example is the only one making use of it right now. I didn't see those errors in the mva demo. Any idea why it runs into those unsupported functions here and not in your demo?

image


CGNonofr:
> @CGNonofr it is all working with the exception of the new `getLocalizationServiceOverride`. Once I configure it, I see a stack of errors. The Statemachine example is the only one making use of it right now. I didn't see those errors in the mva demo. Any idea why it runs into those unsupported functions here and not in your demo?
> 
> image

Probably because the demo registers the multiDiffEditor service override

Are you using the scm service override? because it seems that's the only one using the unsupported method

kaisalmen:
> Are you using the scm service override? because it seems that's the only one using the unsupported method

Not intentionally. The scm service override does not appear in the list of services that is loaded. Adding the multiDiffEditor does not change the behaviour/problem go away.
image

kaisalmen:
@CGNonofr errors are gone, but the `LocalizationOptions` are not used. I tried adding environment and preferences services, but without success.

I noticed that "Configure Display Language" disappeared even from your demo (see comparison with VSCode below). It was there before you fixed my issue. 🙈 
image
image

kaisalmen:
@CGNonofr `getLifecycleServiceOverride` is required for this to work. If you select the language from the menu the restart does not work otherwise. The problem is that the url is changed, but the language is not changed. It always stays English. Manually altering the url and reload has no effect, either.

Is there any specific service required for this to work. I looked again at the demo and the changes you performed and there this just works, but you have all services loaded. What piece of the puzzle am I missing?

kaisalmen:
@CGNonofr nevermind. I found the problem. The loader function is still required and I removed it. 🙈

kaisalmen:
@CGNonofr this is ready for review now. `getLifecycleServiceOverride` is the only service required in addition,

packages/examples/src/utils/localeLoader.ts
```diff
@@ -13,6 +16,36 @@ const localeLoader: Partial<Record<string, () => Promise<void>>> = {
...
+        await import('@codingame/monaco-vscode-language-pack-ru');
+    },
+    tr: async () => {
+        await import('@codingame/monaco-vscode-language-pack-tr');
+    },
+    'zh-hans': async () => {
+        await import('@codingame/monaco-vscode-language-pack-zh-hans');
+    },
+    'zh-hant': async () => {
+        await import('@codingame/monaco-vscode-language-pack-zh-hant');
```
CGNonofr, line 48: didn't you have to install them?

Issue #621, opened 2023-06-07
Unload and reload of react component -> MonacoEditorReactComp ("@typefox/monaco-editor-react")


Stuey61296:
I am working with monaco that will be loaded into multiple different components and connects to a language server. If I were to leave the page where monaco has previously been loaded onto, and then return to the page. I get the console output saying *lifecycleService.js:54 Uncaught (in promise) Error: Lifecycle cannot go backwards*. I am not sure if the instance has not been destroyed properly or there is something I am doing wrong...

My config is below.
![image](https://github.com/TypeFox/monaco-components/assets/53872200/ca7d1698-bb37-4de9-bc2a-b7b0cb3f4521)

My console output is below.
![image](https://github.com/TypeFox/monaco-components/assets/53872200/5dd4dd38-d857-4fff-8b06-fd498565f94e)

kaisalmen:
With #634 I consider this one resolved as well,

PR #634, opened 2024-04-11
monaco-editor-react works in strict mode

### LINKED ISSUE #631: @typefox/monaco-editor-react does not work with React StrictMode enabled
Hi, I was looking to adopt your React wrapper for Monaco to use the language client services. An issue I ran into was, whenever `<StrictMode>` was enabled in React, the editor would fail to load properly and initialize the model. If you remove `<StrictMode>`, it works as expected.

I've reproduced the issue in a [codesandbox for your review/testing.](https://codesandbox.io/p/devbox/typefox-react-strictmode-issue-6q5jr2).

kaisalmen:
Re-Initialization was not properly safe-guarded. This update ensure this is the case. Re-init check functionality was mostly moved to the wrapper as it is unrelated to react, but to user and editor configuration. A simple unit test is introduced as well. The `logger` from the wrapper is exposed, so it can be used by the react-component.

This fixes #631 and #621

### DIFF
--- a/packages/wrapper/test/wrapper.test.ts
+++ b/packages/wrapper/test/wrapper.test.ts
@@ -57,4 +57,20 @@ describe('Test MonacoEditorLanguageClientWrapper', () => {
             await wrapper.initAndStart(config, document.getElementById('monaco-editor-root'));
         }).rejects.toThrowError('init was already performed. Please call dispose first if you want to re-start.');
     });
+
+    test('Test if ', async () => {
+        createMonacoEditorDiv();
+        const wrapper = new MonacoEditorLanguageClientWrapper();
+        const userConfigClassic = createBaseConfig('classic');
+        wrapper.init(userConfigClassic);
+        expect(wrapper.isReInitRequired(userConfigClassic, userConfigClassic)).toBeFalsy();
+
+        const userConfigExtended = createBaseConfig('extended');
+        expect(wrapper.isReInitRequired(userConfigClassic, userConfigExtended)).toBeTruthy();
+
+        const userConfigClassicNew = createBaseConfig('classic');
+        userConfigClassicNew.wrapperConfig.editorAppConfig.useDiffEditor = true;
+
+        expect(wrapper.isReInitRequired(userConfigClassicNew, userConfigClassic)).toBeTruthy();
+    });
 });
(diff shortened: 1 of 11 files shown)


kaisalmen:
In addition to @zzullick in #631 @cdietrich brought this also up a while ago (see #621). I discussed this internally with @sailingKieler and I realized what the root cause is. Good thing is (FMPOV) that the react layer is now even thinner around `monaco-editor-wrapper`.

packages/wrapper/test/wrapper.test.ts
```diff
@@ -57,4 +57,20 @@ describe('Test MonacoEditorLanguageClientWrapper', () => {
             await wrapper.initAndStart(config, document.getElementById('monaco-editor-root'));
         }).rejects.toThrowError('init was already performed. Please call dispose first if you want to re-start.');
     });
+
+    test('Test if ', async () => {
```
CGNonofr, line 61: It look like a good idea to test if, indeed (:

kaisalmen: 😆 oh man thanks, updated ⬇️

CGNonofr (approved):
LGTM!

sailingKieler (approved):
I didn't look into the implementations, yet, but function-wise it looks very good to me.

One kind request: could you make `@typefox/monaco-editor-react` re-export the definitions of `UserConfig`, `WrapperConfig`, `LanguageClientConfig`, etc. ?

Then the import
```
import { MonacoEditorReactComp, type UserConfig } from '@typefox/monaco-editor-react';
```
would be sufficient and I can avoid another one like
```
import type { UserConfig } from "monaco-editor-wrapper";
```
Then I might drop `monaco-editor-wrapper` from my declared dependencies
(with PNPM for example I have to declare each and every dependency explicitly in the package.json)

kaisalmen:
> One kind request: could you make @typefox/monaco-editor-react re-export the definitions of UserConfig, WrapperConfig, LanguageClientConfig, etc. ?

I am no fan of re-exports, because it basically hides where things come from. It should just be the exception and not the rule. People should be very much aware they use "things" from the wrapper as this react component should just be the most thinnest layer possible around `monaco-editor-wrapper` . Is there a better reason than just the need do use another import and package definition? 🙂 

Btw, the `peerDependencies` defined should prevent a wrong combination of underlying packages.

PR #650, opened 2024-05-03
Update code configuration. languageId is no longer required.

### LINKED ISSUE #639: Reconsider languageId usage
The globally configured `languageId` overwrites the `languageId` of any model that is passed via `modelUpdate` to monaco-editor-wrapper (see https://github.com/TypeFox/monaco-languageclient/discussions/638). This is too restrictive and even wrong. We need to revise this approach and eventually make `languageId` no longer mandatory.

kaisalmen:
Re-work how to configure code in `UserConfig` and make `languageId` no longer required and do not enforce it in the editor model.
This is now uri or file extension driven. `languageId` can be optionally enforced (WA for classical editor in the wrapper).

It is now possible to create an editor without any text or model and update the model or `codeResources` later. This is demonstrated here: https://github.com/TypeFox/monaco-languageclient/blob/issue-639/packages/examples/src/langium/statemachine/main.ts#L28-L42

Fixes #639

### DIFF
--- a/packages/examples/src/browser/main.ts
+++ b/packages/examples/src/browser/main.ts
@@ -162,7 +162,7 @@ export const runBrowserEditor = async () => {
 
     await wrapper.start(htmlElement);
 
-    wrapper.getModel()?.onDidChangeContent(() => {
+    wrapper.getTextModel()?.onDidChangeContent(() => {
         validate();
     });
 };
--- a/packages/examples/src/browser/main.ts
+++ b/packages/examples/src/browser/main.ts
@@ -43,9 +43,12 @@ export const runBrowserEditor = async () => {
             },
             editorAppConfig: {
                 $type: 'extended',
-                languageId: 'json',
-                code,
-                codeUri,
+                codeResources: {
+                    main: {
+                        text: code,
+                        uri: codeUri
+                    }
+                },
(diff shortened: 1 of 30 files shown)


kaisalmen:
@CGNonofr are the API changes still ok for a minor version change? WDYT?

CGNonofr:
> @CGNonofr are the API changes still ok for a minor version change? WDYT?

We should keep it simple:
- fix = patch
- feature = minor
- breaking change = major

Is it backward compatible ?

kaisalmen:
> Is it backward compatible ?

The UserConfig is not. So, if you do it by the book it is a major version change. But the change is for the good, because what I did previously was too limiting.

kaisalmen:
@CGNonofr I should add a couple of unit tests for this and I think the new version of wrapper and react comp should be major once.
Do you think this configuration approach for code text / model handeling is an improvement?

CGNonofr (approved):

kaisalmen:
@CGNonofr I just released the new versions on npm. I cleaned-up examples and did some further API alignments directly on main as I realised I missed something. `monaco-languageclient` is a minor change (`8.4.0`). The wrapper advanced to v5 and the react component advanced to v4.

PR #661, opened 2024-05-17
Make code compatible with vite build, small corrections

kaisalmen:
@CGNonofr I realized that the examples don't work well when `vite build` is used.
At some point I want to deploy the examples that don't require a backend to website (e.g. GH Pages).

The need for shx/shelljs is removed. It has been replace by a node script written in TS that is executed with vie-node.


packages/examples/src/langium/langium-dsl/config/extendedConfig.ts
```diff
@@ -10,17 +10,17 @@ import { useOpenEditorStub } from 'monaco-editor-wrapper/vscode/services';
 import { UserConfig } from 'monaco-editor-wrapper';
 import { BrowserMessageReader, BrowserMessageWriter } from 'vscode-languageclient/browser.js';
 import { loadLangiumWorker } from '../wrapperLangium.js';
-import { getTextContent } from '../../../common/client/app-utils.js';
+import langiumLanguageConfig from './langium.configuration.json' assert { type: 'json' };
+import langiumTextmateGrammar from './langium.tmLanguage.json' assert { type: 'json' };
+// @ts-expect-error otherwise the vite notation leads to a compile error
+import text from '../content/example.langium?raw';
```
CGNonofr, line 16: You can create the proper .d.ts instead of ignoring errors

CGNonofr: something like
```typescript
declare module '*?raw' {
  const content: string
  export default content
}
```

kaisalmen: @CGNonofr thank you fixed below. ⬇️

kaisalmen:
@CGNonofr I think this is now ready

CGNonofr (approved):
LGTM

PR #666, opened 2024-05-27
Fix: Multiple Monaco editor issue

### LINKED ISSUE #620: Multiple Monaco editor issue
I want to add multiple Monaco editors on one page. [Demo code](https://github.com/bguo3/stDemo). Seems they share the same objects.
Declare "Aaa" in one editor could be referenced in the other.
image

kaisalmen:
This fixes #620 It actually didn't require a code change. You only create one language client and the other editor re-uses the globally installed for the registered language (= `statemachine` in this case). This may have not been as easily configurable in the past.

In addition the PR contains the update to monaco-vscode-api version `5.2.0`


CGNonofr (approved):
LGTM

wa8f7192:
Thanks for resolving this and updating the examples

cdietrich:
@kaisalmen is there also the option to have the react component twice and have them completely isolated from each other?
(two workers, each editor talking to one?)

kaisalmen:
@cdietrich each editor for a different LS/language?

cdietrich:
for the same language.

see this branch:
https://github.com/cdietrich/my-monaco-editor-react-example/tree/cd/otherFilesExperiments_createModelReference_multi-newer-componentTwice

currently both workers seem to see the files from the other.

CGNonofr:
I don't think it will be able, monaco/VSCode don't support having multiple isolated "instances" in parallel

The closer you can get is having 2 workspace folders. You can configure a language client to match only a single workspace folder.

kaisalmen:
> The closer you can get is having 2 workspace folders. You can configure a language client to match only a single workspace folder.

We should add that to our Troubleshooting section

cdietrich:
hmm ok. so this is why monaco dispatches all events to all clients.
as we have tabs inside our react app, we are looking into how can unmount when tab is not visible

CGNonofr:
An alternative can be to use Iframes also of course

kaisalmen:
@cdietrich and @CGNonofr btw, you can use MessageChannels to communicate in between IFrames or to channel the communication between main and worker or between worker. 

Idea: You could have logic either on client or the worker or both to disable communication-

kaisalmen:
@cdietrich disconnecting and re-connecting the languageclient on tab change is not an option? You don't have to dispose the worker. This should be fast.

cdietrich:
we need to check

kaisalmen:
If something is missing, let us know.

PR #669, opened 2024-06-03
chore: localeLoader: allow manually select which locale to load

undefined-moe:


packages/wrapper/src/vscode/localeLoader.ts
```diff
@@ -59,3 +60,5 @@ export const initLocalLoader = async () => {
         }
     }
 };
+/** @deprecated use `initLocaleLoader` instead */
+export const initLocalLoader = initLocaleLoader;
```
kaisalmen, line 64: Thank you for spotting the typo in the function name. There is no need to deprecate the old name, just remove it.

CGNonofr (commented):
Looks like a good idea indeed!

kaisalmen (changes requested):
Thank you very much. No need to deprecate the wrong function name. Just drop that and then I will approve this. 🎉

packages/wrapper/src/vscode/localeLoader.ts
undefined-moe: I've got that updated 👍

kaisalmen (approved):
LGTM 🎉

PR #670, opened 2024-06-03
Wrapper: Make codeResources and useDiffEditor optional in EditorAppConfig

kaisalmen:
With latest changes it is now also possible to remove any `codeResources` or `useDiffEditor` from the editor app configuration.

Dependencies were updated as well


packages/wrapper/src/editorAppBase.ts
```diff
@@ -109,7 +109,7 @@ export abstract class EditorAppBase {
     }
 
     async createEditors(container: HTMLElement): Promise<void> {
-        if (this.getConfig().useDiffEditor) {
+        if (this.getConfig().useDiffEditor === true) {
```
CGNonofr, line 112: nitpicking: it seems what you want here is a default value for the `useDiffEditor` flag, using `=== true` for it is confusing to me because the default value doesn't appear anywhere. I would prefer `if (this.getConfig().useDiffEditor ?? false) {`

packages/wrapper/test/editorAppClassic.test.ts
```diff
@@ -49,6 +49,10 @@ describe('Test EditorAppClassic', () => {
         const app = new EditorAppClassic('test', createBaseConfig('classic'));
         expect(app.isAppConfigDifferent(orgConfig, config, false)).toBeFalsy();
 
+        if (!config.codeResources) {
+            config.codeResources = {};
+        }
```
CGNonofr, line 54: nitpicking, you can use `config.codeResources ??= {}`

btw, I've noticed you don't seem to use the `strict-boolean-expressions` eslint rule. It seems import to code quality and reliability, is there a reason for not using it?

packages/wrapper/src/editorAppBase.ts
kaisalmen: Good point, `??` syntax is more readable. In the end both statements lead to the same result.

packages/wrapper/test/editorAppClassic.test.ts
kaisalmen: > nitpicking, you can use config.codeResources ??= {}

I didn't really use it like this before. Thanks for the hint.

> btw, I've noticed you don't seem to use the `strict-boolean-expressions` eslint rule. It seems import to code quality and reliability, is there a reason for not using it?

Uih, I just integrated that and I wasn't aware of the issues it revealed. The checking is taken from Langium. After seeing the effects of this rule I will recommend to enable it there as well.

kaisalmen: @CGNonofr btw, have you ported any configuration to eslint 9 already? It seems to be not very straight forward. I did not really spent time to investigate / understand it, though.

CGNonofr: No, eslint is the only package not up to date on the monaco-vscode-api side, I didn't investigate either, everything seems to broke, I will probably wait a little bit for all the plugin to update

packages/wrapper/src/editorAppBase.ts
kaisalmen: Done ⬇️

packages/wrapper/test/editorAppClassic.test.ts
kaisalmen: Done ⬇️

packages/client/src/index.ts
```diff
@@ -20,7 +20,7 @@ export class MonacoLanguageClient extends BaseLanguageClient {
     protected readonly connectionProvider: IConnectionProvider;
 
     constructor({ id, name, clientOptions, connectionProvider }: MonacoLanguageClientOptions) {
-        super(id || name.toLowerCase(), name, clientOptions);
+        super((id ?? 'unknown-id') || name.toLowerCase(), name, clientOptions);
```
CGNonofr, line 23: I'm confused by the double usage of `??` and `||`

kaisalmen: My mistake. What confuses me now is that eslint finds problems I don't see locally 😐

kaisalmen: Done

.github/workflows/actions.yml
```diff
@@ -18,10 +18,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v4
-    - name: Use Node.js
-      uses: actions/setup-node@v4
-      with:
-        node-version: '20'
+    - name: Volta
+      uses: volta-cli/action@v4
```
CGNonofr, line 22: Oh nice, didn't know about it

packages/client/src/index.ts
CGNonofr: It seems to be still failing :thinking:

kaisalmen:
@CGNonofr eslint produces errors I don't see on Windows, but only on Linux. I can reproduce this locally.
I added `volta-cli/action@v4` to enforce volta usage in the pipeline.

kaisalmen:
@CGNonofr I got it. Code from the dependent packages in the repo are not available. Type resolution is therefore not correct. The code needs to be built first.

.github/workflows/actions.yml
kaisalmen: Me neither, I just found it

kaisalmen:
@CGNonofr problems are resolved. This is ready now.

CGNonofr (approved):
LGTM, thanks for the changes :)

PR #672, opened 2024-06-04
Move localeLoader to monaco-languageclient

kaisalmen:
It makes more sense there, because this is not a specific functionality of the wrapper and could then be used without it. WDYT?


CGNonofr:
I'm not sure about the exact architectures of the projects, it doesn't seem really related to the languageclient package to me

kaisalmen:
> I'm not sure about the exact architectures of the projects, it doesn't seem really related to the languageclient package to me

It is more a utility function for `monaco-vscode-api`, but you don't need the wrapper for it. But, if you think it should not go to the client, I will leave it where it currently is and close this PR without merging.

CGNonofr (approved):
Change is simple, so I'll let you decide what you think is the best

kaisalmen:
@CGNonofr I changed my mind and let it stay where it is.

PR #680, opened 2024-06-22
Dispose extension files in EditorAppExtended

### LINKED ISSUE #679: wrapper.dispose() does not get rid of extension
I assume that `MonacoEditorLanguageClientWrapper.dispose()` is supposed unregister anything that it registered during `init` and `start`, right? Then it does not seem to work correctly with extensions that use `filesOrContents`.

Here is a small example (full reproduction repo here: https://github.com/abentkamp/monacotest3/tree/reinit) :
```ts
import { MonacoEditorLanguageClientWrapper, UserConfig } from 'monaco-editor-wrapper';

const extensionFilesOrContents = new Map<string, string | URL>();
extensionFilesOrContents.set('/test.json', "Test");

const userConfig : UserConfig = {
  wrapperConfig: {
    editorAppConfig: {
      $type: 'extended',
      codeResources: {
        main: {
          text: 'print("Hello, World!")',
          uri: '/workspace/hello.py'
        }
      },
      extensions: [{
        config: {
            name: 'myextension',
            publisher: 'me'
        },
        filesOrContents: extensionFilesOrContents
      }]
    }
  }
};

const run = async () => {
  const htmlElement = document.getElementById('monaco-editor-root');

  const wrapper = new MonacoEditorLanguageClientWrapper();
  await wrapper.initAndStart(userConfig, htmlElement);
  await wrapper.dispose();

  const wrapper2 = new MonacoEditorLanguageClientWrapper();
  await wrapper2.initAndStart(userConfig, htmlElement);
  await wrapper2.dispose();
}

run()
```

I get the following error:
```
Uncaught (in promise) Error: file 'extension-file://me.myextension/' already exists
    registerFile files.js:307
    registerExtensionFile files.js:762
    registerExtensionFileUrl extensions.js:29
    registerFileUrl extensions.js:122
    init editorAppExtended.ts:93
    start wrapper.ts:100
    initAndStart wrapper.ts:85
    run main.ts:36
    async* main.ts:40
[files.js:307:18](http://localhost:5174/node_modules/@codingame/monaco-vscode-files-service-override/files.js)
    run main.ts:38
    InterpretGeneratorResume self-hosted:1469
    AsyncFunctionThrow self-hosted:856
    (Async: async)
    <anonymous> main.ts:40
```

vrama628:
Fixes #679 

There was a bug where disposing and then reinitializing an `EditorAppExtended` where an extension has `filesOrContents` would result in an exception being thrown due to not disposing these files from the last time the extension was registered. This fixes that by disposing of the results of `registerFileUrl` in the `disposeApp` method.

I added a test to exercise this; the test introduced in the first commit exhibits the `Error: file 'extension-file://undefined.undefined/' already exists` error, and that error goes away in the second commit. Unfortunately, when I run this test suite, I also get a bunch of "Array buffer allocation failed" errors that seem to come from somewhere in VSCode's services. I tried using the `updateExtendedAppPrototyp` test helper to fix this, but that seems to remove so much functionality that the original "file already exists" error no longer happens in the first place. Open to suggestions on how to fix the test, or I can just remove the test so we can ship the fix.

### DIFF
--- a/packages/wrapper/src/editorAppExtended.ts
+++ b/packages/wrapper/src/editorAppExtended.ts
@@ -10,7 +10,7 @@ import { registerExtension, IExtensionManifest, ExtensionHostKind } from 'vscode
 import { Logger } from 'monaco-languageclient/tools';
 import { UserConfig } from './userConfig.js';
 import { verifyUrlOrCreateDataUrl, ModelUpdateType, isEqual, isModelUpdateRequired } from './utils.js';
-import { DisposableCollection } from 'vscode-ws-jsonrpc';
+import { DisposableStore } from 'vscode/monaco';
 
 export type ExtensionConfig = {
     config: IExtensionManifest | object;
@@ -49,7 +49,7 @@ export class EditorAppExtended extends EditorAppBase {
 
     private config: EditorAppConfigExtended;
     private extensionRegisterResults: Map<string, RegisterLocalProcessExtensionResult | RegisterExtensionResult | undefined> = new Map();
-    private subscriptions: DisposableCollection = new DisposableCollection();
+    private subscriptions: DisposableStore = new DisposableStore();
 
     constructor(id: string, userConfig: UserConfig, logger?: Logger) {
         super(id);
@@ -93,7 +93,7 @@ export class EditorAppExtended extends EditorAppBase {
                 if (extensionConfig.filesOrContents && Object.hasOwn(extRegResult, 'registerFileUrl')) {
                     for (const entry of extensionConfig.filesOrContents) {
                         const registerFileUrlResult = (extRegResult as RegisterLocalExtensionResult).registerFileUrl(entry[0], verifyUrlOrCreateDataUrl(entry[1]));
-                        this.subscriptions.push(registerFileUrlResult);
+                        this.subscriptions.add(registerFileUrlResult);
                     }
                 }
                 allPromises.push(extRegResult.whenReady());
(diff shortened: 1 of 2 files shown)


packages/wrapper/src/editorAppExtended.ts
```diff
@@ -106,6 +108,8 @@ export class EditorAppExtended extends EditorAppBase {
     disposeApp(): void {
         this.disposeEditors();
         this.extensionRegisterResults.forEach((k) => k?.dispose());
+        this.subscriptions.forEach((k) => k.dispose());
+        this.subscriptions.length = 0;
```
CGNonofr, line 112: nit: for informations (@kaisalmen ) `vcode/monaco` export the `DisposableCollection` class from VSCode, do not hesitate to use it :)

vrama628: Thanks, I've updated it to use `DisposableCollection`. I wasn't able to find it in the `vscode` or `monaco-editor` packages, but I did find it in the `vscode-ws-jsonrpc` package so I used it from there.

kaisalmen: @CGNonofr you meant `import { DisposableStore } from 'vscode/monaco';` like used here https://github.com/CodinGame/monaco-vscode-api/blob/main/src/service-override/tools/views.ts#L258-L367, right?

vrama628:
@kaisalmen Thanks, I've moved the test into a separate draft PR https://github.com/TypeFox/monaco-languageclient/pull/688 and updated to `import { DisposableStore } from 'vscode/monaco'`.

kaisalmen (approved):
LGTM. Thank you.

PR #682, opened 2024-06-22
Wait for Monaco to initialize when MonacoEditorReactComp props are updated

### LINKED ISSUE #622: Introduce unit tests for monaco-editor-react
We already introduced multiple unit tests for monaco-editor-wrapper, but don't have any for monaco-editor-react.
This is not as straightforward as initially thought. It is likely the combination of vitest with browser mode and react. I already started some prototyping on https://github.com/TypeFox/monaco-components/tree/vitest-react

### LINKED ISSUE #681: MonacoEditorReactComp throws "You cannot update the code resources as neither editor or diff editor is available."
When the `MonacoEditorReactComp` component is rerendered before Monaco has finished initializing, it throws errors like `TypeError: Cannot read properties of undefined (reading 'isAppConfigDifferent')` and `Error: You cannot update the code resources as neither editor or diff editor is available.`

vrama628:
Fixes #681 and fixes #622

In `MonacoEditorReactComp` there's a race condition between the Monaco editor initializing and the component rerendering; if a component rerenders and triggers `componentDidUpdate` before Monaco has finished initializing, errors happen as described in the linked issue.

The first commit introduces a test that exhibits these errors, and the second commit fixes it by awaiting `this.isRestarting` in `componentDidUpdate`.


package.json
```diff
@@ -5,25 +5,27 @@
     "@codingame/esbuild-import-meta-url-plugin": "~1.0.2",
     "@codingame/monaco-vscode-rollup-vsix-plugin": "~5.2.0",
     "@rollup/pluginutils": "~5.1.0",
+    "@testing-library/react": "^16.0.0",
```
kaisalmen, line 8: Please use `~` instead of `^`, so only minor version updates are used.

package.json
```diff
@@ -5,25 +5,27 @@
...
     "@types/node": "~20.14.1",
     "@types/react": "~18.3.3",
     "@types/react-dom": "~18.3.0",
     "@types/vscode": "~1.89.0",
     "@typescript-eslint/eslint-plugin": "~7.12.0",
     "@typescript-eslint/parser": "~7.12.0",
+    "@vitejs/plugin-react": "^4.3.1",
```
kaisalmen, line 15: Please use `~` instead of `^`, so only minor version updates are used.

kaisalmen (changes requested):
Thank you very much! Only two minor tweaks required. I tested this successfully already and it is basically ready for merge. Btw, you closed issue #622 with this PR as well. 🎉

kaisalmen (approved):
Thank you. LGTM.

PR #696, opened 2024-07-02
Fix Diff Editor Model Refs

zgbpr:
# Summary

Fixes a bug that caused both diff editors to display the same model.

# Details

We ran into a bug with the diff editor where both the modified editor and original editor displayed the same content. When we inspected the editors in the browser, we found that they had the same `data-uri` value.

After tracing through the `editorAppBase.ts` file, we saw that the `getEditorUri` function accepts a boolean to determine whether to append 'Original'. But the `buildModelRef` function always passed `false` when building the model reference for both editors.

This pull request fixes the bug by passing a boolean parameter from `buildModelRef` to `getEditorUri` and ensuring that the URIs are different for both editors.

# Screenshots

Here is a screenshot before the change, showing how the modified editor and original editor displayed the same content.

Screenshot 2024-07-01 at 7 51 16 PM

Here is a screenshot after the change, showing how the modified editor and original editor now display different content.

Screenshot 2024-07-01 at 7 52 32 PM

### DIFF
--- a/packages/wrapper/src/editorAppBase.ts
+++ b/packages/wrapper/src/editorAppBase.ts
@@ -172,8 +172,8 @@ export abstract class EditorAppBase {
     }
 
     async buildModelRefs(codeResources?: CodeResources): Promise<ModelRefs> {
-        const modelRef = await this.buildModelRef(codeResources?.main);
-        const modelRefOriginal = await this.buildModelRef(codeResources?.original);
+        const modelRef = await this.buildModelRef(codeResources?.main, false);
+        const modelRefOriginal = await this.buildModelRef(codeResources?.original, true);
 
         return {
             modelRef,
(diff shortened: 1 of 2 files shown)


kaisalmen:
@zgbpr thank you for spotting this and providing the fix. Can you please add a unit test? https://github.com/TypeFox/monaco-languageclient/blob/main/packages/wrapper/test/wrapper.test.ts#L114-L133 should be a good template to derive a new test from.

kaisalmen (approved):
> Let me know if you'd prefer to verify the names in a separate test.

No, that is fine. Thank you. LGTM

zgbpr:
Thanks @kaisalmen! Do you happen to know when these changes might be included with a release?

kaisalmen:
@zgbpr here you go: https://www.npmjs.com/package/@typefox/monaco-editor-react/v/4.3.1

zgbpr:
Amazing, thank you! 🙏

zgbpr:
Hmm, I pulled down v4.3.1 of `@typefox/monaco-editor-react` but it looks like the underlying version of `monaco-editor-wrapper` still contains the diff model ref bug.

Any chance that there needs to be a v5.3.1 release of `monaco-editor-wrapper` as well?

kaisalmen:
Sorry, my fault. New versions will come soon.

kaisalmen:
@zgbpr her you go:
https://www.npmjs.com/package/monaco-editor-wrapper/v/5.3.1
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/4.3.2

zgbpr:
Thank you for the quick updates, @kaisalmen! That's working well. 🎉

PR #707, opened 2024-07-11
Update to monaco-vscode-api 7.0.3

kaisalmen:
All dependencies (except eslint) and versions have been bumped.
Good news: the browser mode enhancements of vitest v2 allow to use the wrapper as is. No more need to drop services


CGNonofr (approved):

kaisalmen:
@CGNonofr that all test now run without tweaks in vitest browser mode is really good news. 👍

PR #714, opened 2024-07-22
Open default workspace

abentkamp:
Currently, the `configureServices` function creates a default workspace config, but it is not passed to the `initialize` function. This PR passes the configuration to the `initialize` function. Moreover, to work properly, the `workspaceUri` must have the extension `.code-workspace`.

I previously reported a related issue here: https://github.com/CodinGame/monaco-vscode-api/issues/470


kaisalmen (approved):
Thank you! LGTM

PR #734, opened 2024-08-23
Add Clangd-wasm example, build LS images with GHA

kaisalmen:
This PR adds the clangd wasm port from https://github.com/guyutongxue/clangd-in-browser into a new example. The client side of the example is already clean-up / aligned to the other example. **TODO**: The server side code needs some polishing.

In addition the following things have been added along:
- GHA actions that build all container based language servers (groovy, eclipse.jdt.ls, clangd). They need to be manually started, but images are stored in the container registry already

Once we have clangd in integrated, we have together with pyright and eclispe.jdt.ls enough complex language server  (worker and web socket conn.) available to integrate different ways of file synchronizations. This PR enables more future enhancements.


CGNonofr:
I did some investigation around clangd and the possibility to run it in the browser already, here's what I've found:
- there are some official investigation on how to make it fully support WASI out of the box: https://github.com/llvm/llvm-project/pull/92677
- VSCode has an implementation of the WASI protocol in the extension host using https://marketplace.visualstudio.com/items?itemName=ms-vscode.wasm-wasi-core

I was able to run a small C program compiled into wasm, properly accessing the VSCode virtual filesystem. But since it doesn't support some basic features like threads, it failed to run any "advanced" binaries

The only option currently seems to be https://github.com/guyutongxue/clangd-in-browser, but I wonder if it's able to interact with the filesystem, and how?

kaisalmen:
> I did some investigation around clangd and the possibility to run it in the browser already

Great, thanks for sharing 👍 If this becomes fully working in the browser at one point in time (via whatever tech) it will be awesome.

> The only option currently seems to be https://github.com/guyutongxue/clangd-in-browser, but I wonder if it's able to interact with the filesystem, and how?

It brings its own virtual file system. You can read and write files (done here: https://github.com/TypeFox/monaco-languageclient/blob/clangd-example/packages/examples/src/clangd/worker/clangd-server.ts)

CGNonofr:
> It brings its own virtual file system. You can read and write files (done here: https://github.com/TypeFox/monaco-languageclient/blob/clangd-example/packages/examples/src/clangd/worker/clangd-server.ts)

Isn't there any way to provide our own implementation instead? which would access the VSCode filesystem? That would be a shame to have to watch the filesystem to register the files one by one

kaisalmen:
> Isn't there any way to provide our own implementation instead?

The wasm already contains the header files meaning when you have it loaded you can read all the header files as far as I understand it. Something must be adjusted in the build process to change that, I guess. I did not investigate any of this so far.

kaisalmen:
But now we can start to play with it and try things...

kaisalmen:
@CGNonofr I will open up a second PR with the wrapper improvements and base this one on it. I need to fix some build problems first.

kaisalmen:
@CGNonofr maybe we should this merge at some, even if there are improvements required and improve it subsequently (file handling needs polishing and even some parts should be extracted out the examples. I have to make up my mind)

The example should contain a sentence that this is still a prototype and the underlying code is still evolving.

packages/client/src/fs/endpoints/defaultEndpoint.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    getEndpointType(): EndpointType {
+        return this.endpointType;
+    }
+
+    readFile(params: FileReadRequest): Promise<FileReadRequestResult> {
+        this.logger?.info(`Reading file: ${params.resourceUri}`);
+        return Promise.resolve({
+            status: 'denied',
```
CGNonofr, line 35: any reason why it's not throwing an error?

CGNonofr (approved):

packages/client/src/fs/endpoints/defaultEndpoint.ts
kaisalmen: Because the status carries the info. It could be a `Promise.reject`, too. The whole file system / file handling realm will still evolve, so if it is not critical I will leave it as is.

kaisalmen:
@CGNonofr this is ready for merge on my side. There will be improvements.

PR #739, opened 2024-08-27
Turned react component from class to function

jbicker:
As the title states, the react component became a function component to respect modern react patterns.


kaisalmen:
@jbicker I have updated the test as discussed offline. Please have a look.
I also moved `isReInitRequired` to utils and did not make it deprecated for now.

PR #749, opened 2024-09-23
Update to monaco-vscode-api 9.0.x

kaisalmen:
@CGNonofr I have a weird observation. When I use monach `TextEditorWorker` is loaded and working and there is no error in the console.
When I use textmate, first I now need to specify the `TextMateWorker` otherwise there are errors printed (first screenshot). That wasn't required before. Once I do specify it, I see the warning in the second screenshot. The worker iscreated in the factory, but then something goes wrong.  It can be reproduced with the langium example. It can be loaded with both configurations. I don't see this beahviour in the monaco-vscode-api demo locally. Do you have any idea?

No `TextMateWorker` provided:
image

`TextMateWorker` provided:
image


CGNonofr:
> When I use textmate, first I now need to specify the `TextMateWorker` otherwise there are errors printed (first screenshot). That wasn't required before.

Indeed, I've missed that change because it wasn't announced: there is a `editor.experimental.asyncTokenization` settings which allows to switch between local and worker syntax highlighting. That settings was false by default and was [switched on recently](https://github.com/microsoft/vscode/pull/226410)

> Once I do specify it, I see the warning in the second screenshot. The worker iscreated in the factory, but then something goes wrong. It can be reproduced with the langium example. It can be loaded with both configurations. I don't see this beahviour in the monaco-vscode-api demo locally. Do you have any idea?

Are you aware of the [breaking changes of the v9.0.0](https://github.com/CodinGame/monaco-vscode-api/releases/tag/v9.0.0) ?

kaisalmen:
> Are you aware of the [breaking changes of the v9.0.0](https://github.com/CodinGame/monaco-vscode-api/releases/tag/v9.0.0) ? the error you get seems to be

Yes, the names are adjusted. That error above is not thrown. There is something async going on. When I halt in the debugger in the factory, it does not occur.

> Indeed, I've missed that change because it wasn't announced: there is a editor.experimental.asyncTokenization 

The settings are applied too late. That is different in your demo. I have to adjust.

kaisalmen:
@CGNonofr let's defer the resolution of this problem (I could create a new issue) and merge this PR (after review, of course). Especially "unplanned" enhancements to logging, worker factory and start order make this PR very useful apart from the mva version upgrade.

packages/wrapper/src/editorAppClassic.ts
```diff
@@ -44,8 +45,17 @@ export class EditorAppClassic extends EditorAppBase {
         };
     }
 
+    override async loadUserConfiguration() {
+        if (this.config.editorOptions?.['semanticHighlighting.enabled'] !== undefined) {
+            // use updateConfiguration here as otherwise semantic highlighting will not work
+            await updateUserConfiguration(JSON.stringify({
+                'editor.semanticHighlighting.enabled': this.config.editorOptions['semanticHighlighting.enabled']
+            }));
```
CGNonofr, line 53: Is it still required? as the default value is "configuredByTheme"

also, doing that will override any existing user configuration, better using `StandaloneServices.get(IConfigurationService).updateValue('editor.semanticHighlighting.enabled', true)` instead maybe

packages/client/src/tools/index.ts
```diff
@@ -3,42 +3,5 @@
...
-        }
-    }
-
-    debug(message: string, force?: boolean) {
-        if (this.enabled && (this.debugEnabled || force === true)) {
-            console.debug(message);
-        }
-    }
-}
+export type * from './logging.js';
```
CGNonofr, line 6: is there a point of exporting types it you already export everything?

packages/client/src/tools/logging.ts
```diff
@@ -0,0 +1,66 @@
...
+        case 4:
+            return LogLevel.Warning;
+        case 'Error':
+        case 'error':
+        case '5':
+        case 5:
+            return LogLevel.Error;
+        default:
+            return LogLevel.Off;
+    }
```
CGNonofr, line 65: why do we need to handle all these values ? (:

packages/client/src/vscode/services.ts
```diff
@@ -105,21 +97,19 @@ export const initServices = async (instruction: InitServicesInstruction) => {
...
-    reportServiceLoading(userServices, instruction.logger);
+    reportServiceLoading(userServices, instructions.logger);
 
-    if (instruction.performChecks === undefined || instruction.performChecks()) {
-        await initialize(userServices, undefined, lc.workspaceConfig);
-        const logLevel = lc.workspaceConfig?.developmentOptions?.logLevel ?? LogLevel.Info;
-        StandaloneServices.get(ILogService).setLevel(logLevel);
+    if (instructions.performChecks === undefined || (typeof instructions.performChecks === 'function' && instructions.performChecks())) {
+        await initialize(userServices, undefined, instructions.workspaceConfig);
+        StandaloneServices.get(ILogService).setLevel(instructions.logger?.getLevel() ?? LogLevel.Off);
```
CGNonofr, line 112: FYI, you can set the log level [this way](https://github.com/CodinGame/monaco-vscode-api/blob/61752f5a8e77b1359c8648f8455d048d483c7e79/demo/src/setup.common.ts#L241-L243)

packages/client/src/tools/index.ts
kaisalmen: Is `export *` including types? Oh, yeah it seems to be the case. I will remove it everywhere.

packages/client/src/tools/logging.ts
kaisalmen: I don't get it. It's just a helper function.

packages/client/src/vscode/services.ts
kaisalmen: Yes thanks, I need to updated the config if it is not present. I want to have this visible on higher level in the wrapperConfig. I will adapt the code.

packages/client/src/tools/logging.ts
CGNonofr: Yes, I'm not sure why we need a function that takes anything as parameter and tries its best to understand what it's about, instead of having strong types?

kaisalmen: I don't have to `import { LogLevel } from 'vscode/services';` in all examples, but maybe you are right. It makes sense when you want to parse command-line values, but we don't need to do this here.

packages/wrapper/src/editorAppClassic.ts
kaisalmen: Done ⬇️

packages/client/src/tools/index.ts
kaisalmen: Done ⬇️

packages/client/src/tools/logging.ts
kaisalmen: Done ⬇️

packages/client/src/vscode/services.ts
kaisalmen: Done ⬇️

kaisalmen:
@CGNonofr this is ready. Any ideas regarding my findings: https://github.com/TypeFox/monaco-languageclient/pull/749#issuecomment-2371750627

kaisalmen:
You can reproduce this by simply doing it manually. Something must happen in the worker and then the error is thrown, but the content does not help me a least to understand why it dies:
```javascript
const worker = new Worker(new URL('@codingame/monaco-vscode-textmate-service-override/worker', import.meta.url), { type: 'module' });
worker.addEventListener('error', (e) => {
    console.error(e);
});
```

kaisalmen:
I tried to debug `textMateTokenizationWorker.worker`, but it does not halt on any breakpoints I tried. I see it created in the browser and when I halt in the callback it is still visible, but after basically already terminated. When you don't have that breakpoint it is gone immediately after it was instantiated.  I have no experience in the inner workings of the `SimpleWorkerServer` / `SimpleWorkerProtocol`

packages/wrapper/src/vscode/services.ts
```diff
@@ -50,7 +47,20 @@ export const configureServices = async (config: VscodeServicesConfig): Promise<I
             }
         };
     }
-    mergeServices(config.specificServices ?? {}, serviceConfig.userServices);
+    mergeServices(config.specificServices, serviceConfig.userServices);
+
+    // set the log-level via the development settings
+    const devLogLevel = serviceConfig.workspaceConfig.developmentOptions?.logLevel;
+    if (serviceConfig.workspaceConfig.developmentOptions?.logLevel === undefined) {
```
CGNonofr, line 54: you probably mean `devLogLevel` instead of `serviceConfig.workspaceConfig.developmentOptions?.logLevel`

packages/wrapper/src/vscode/services.ts
```diff
@@ -50,7 +47,20 @@ export const configureServices = async (config: VscodeServicesConfig): Promise<I
...
+
+        // this needs to be done so complicated, because developmentOptions is read-only
+        const devOptions = (serviceConfig.workspaceConfig.developmentOptions as unknown as Record<string, unknown> | undefined) ?? {};
```
CGNonofr, line 57: what about 

```
const devOptions = {
  ...serviceConfig.workspaceConfig.developmentOptions
}
```
?

packages/wrapper/src/vscode/services.ts
kaisalmen: oh, yes, I forgot to replace that 😅

packages/wrapper/src/vscode/services.ts
kaisalmen: Thank you. It can be written as such now:
```javascript
const devOptions: Record<string, unknown> = {
    ...serviceConfig.workspaceConfig.developmentOptions
};
devOptions.logLevel = config.logLevel;
(serviceConfig.workspaceConfig.developmentOptions as Record<string, unknown>) = Object.assign({}, devOptions);
```

packages/wrapper/src/vscode/services.ts
kaisalmen: Done ⬇️

packages/wrapper/src/vscode/services.ts
kaisalmen: Done ⬇️

kaisalmen:
> Please tell me how to reproduce the issue like I am 5yo (:

Sorry, it is too late for this 😆 

> Do you want me to have a look?

Yes, please have a look and reproduce it first and tell me I am not insane. It works in your demo. Something odd is going on here. You can use the Langium-DSL example and flip [here](https://github.com/TypeFox/monaco-languageclient/blob/mva-next/packages/examples/src/langium/langium-dsl/config/extendedConfig.ts#L76) or just the code snippet from https://github.com/TypeFox/monaco-languageclient/pull/749#issuecomment-2373776926 That is sufficient to reproduce.

kaisalmen:
It dies right after this:
image

kaisalmen:
@CGNonofr let's merge this. If you have any further idea regarding the issue above let's just write it down here or continue in a new issue. Both is fine with me.

One difference between your demo and the usage here is that you reference local packages/ locally build code and here the npm packages are used. That should no make a difference, but it is the only obvious difference.

CGNonofr:
I'm not sure to understand, you want to merge something that's not fully working?

kaisalmen:
When the Textmate Worker is disabled, then it is working fine. I could force that temporarily. WDYT?
I want those other changes in main, so I can continue with other work and release a next version.

We should fix/understand/find a solution for the issue before performing the next major version release.

CGNonofr (approved):

CGNonofr:
@kaisalmen the solution is to add `vscode-textmate` and `vscode-oniguruma` in the vite config in `optimizeDeps.include`

The issue is that those libraries are still published in commonjs andthe worker is configured as type: module I think

CGNonofr:
https://github.com/microsoft/vscode-textmate/pull/240 :shrug:

CGNonofr:
btw, I didn't find any way to debug workers in any browser, that's really a pain in the *ss. You just received an "Error" without any explanation nor details, while the error is that the worker code is not valid

kaisalmen:
@CGNonofr awesome. It works. Thank you very much! How long did it take you to find out? Can I buy you a beer or something equivalent? 😆 
image

> btw, I didn't find any way to debug workers in any browse

Once they are loaded (that is sometimes like here the nasty point) you can add debug points and that works with Chrome. If you use a VSCode launch config that works (also in this project, btw).

CGNonofr:
> @CGNonofr awesome. It works. Thank you very much! How long did it take you to find out? Can I buy you a beer or something equivalent? 😆

No worries! :) It didn't really took my long because I already had similar problem in the monaco-vscode-api demo some time ago

cdietrich-everest:
@kaisalmen do we have a release eta for this one

cdietrich:
@kaisalmen do we have a release eta for this one

kaisalmen:
@cdietrich there are a pre-releases available (latest main commit):
https://www.npmjs.com/package/monaco-languageclient/v/9.0.0-next.1
https://www.npmjs.com/package/monaco-editor-wrapper/v/6.0.0-next.1
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/6.0.0-next.1

No fixed ETA, yet

cdietrich:
nice. does this one ring a bell
grafik

PR #756, opened 2024-10-04
Support all arguments for monaco-vscode-api initialize

kaisalmen:
Support all arguments for monaco-vscode-api initialize implements https://github.com/CodinGame/monaco-vscode-api/issues/371

This also allows to configure editor-, view- or workspace-service. This is a preparation for further enhancements.
The `htmlContainer` is now configured at a single place and has become mandatory. This is also a preparation for future enhancements.
@CGNonofr I needed to bring back the workaround for semantic highlighting when the classic/monarch config is used. This time the config is augmented.


packages/wrapper/src/editorAppClassic.ts
```diff
@@ -72,6 +77,20 @@ export class EditorAppClassic extends EditorAppBase {
...
+            try {
+                userConfig = JSON.parse(await getUserConfiguration());
+            } catch (e) {
+                userConfig = {};
+            }
+            const json = JSON.stringify({
+                ...userConfig,
+                'editor.semanticHighlighting.enabled': this.config.editorOptions['semanticHighlighting.enabled']
+            });
+            await updateUserConfiguration(json);
```
CGNonofr, line 92: I'm not sure to understand why this is needed?

btw, you still reformat the file, while using `configurationService.updateValue` would not

kaisalmen: > I'm not sure to understand why this is needed?

If you set it via the editor options https://github.com/TypeFox/monaco-languageclient/blob/main/packages/examples/src/langium/langium-dsl/config/classicConfig.ts#L39-L43 the semanticHighlighting option is not applied. That's the reason why I brought back the workaround.

> btw, you still reformat the file, while using configurationService.updateValue would not

I don't understand what you mean

CGNonofr: You're parsing the json file, add a field then serialize it (and without formatting in single line)

VSCode includes method to update the user configuration without breaking the formatting of the configuration file (it just update or add the new field keeping the formatting of the rest of the file)

kaisalmen: Ok, now I understand. I just pushed an update.

packages/wrapper/src/editorAppClassic.ts
```diff
@@ -79,18 +79,10 @@ export class EditorAppClassic extends EditorAppBase {
...
-            } catch (e) {
-                userConfig = {};
-            }
-            const json = JSON.stringify({
-                ...userConfig,
-                'editor.semanticHighlighting.enabled': this.config.editorOptions['semanticHighlighting.enabled']
-            });
-            await updateUserConfiguration(json);
+            const editorConfig = workspace.getConfiguration('editor.semanticHighlighting');
+            await editorConfig.update('enabled', this.config.editorOptions['semanticHighlighting.enabled'], ConfigurationTarget.Workspace);
```
CGNonofr, line 83: You now update the workspace configuration instead of the user configuration, is it in purpose?

Note that you can also chose to update the memory configuration (there are many layers of configuration that are merged)

kaisalmen: Is `ConfigurationTarget.WorkspaceFolder` (3) = user configuration ?

CGNonofr: You are setting the configuration via the Extension API while I was suggesting to set it via the service directly. Via the api you don't have access to all the targets

kaisalmen: But the service does not expose those functions or does it?

CGNonofr: What do you mean? you can use
```typescript
import { ConfigurationTarget, IConfigurationService, StandaloneServices } from 'vscode/services'

StandaloneServices.get(IConfigurationService).updateValue('editor.semanticHighlighting', 'enabled', ConfigurationTarget.MEMORY)
```

kaisalmen: Thank you, I was wondering if it is exposed via `@codingame/monaco-vscode-configuration-service-override`.

kaisalmen: Done ⬇️

CGNonofr: So you've chosen USER over MEMORY? :)

kaisalmen: Yes, I thought it makes sense. It is anyway only for the classical config. These days, I would recommend to use the extended mode directly. It could make sense to drop the other mode at some point in the future.

kaisalmen: There is only one example left that uses the `classic` / monarch app config.

CGNonofr: I was thinking... you can provide a default configuration to the initialize method also, see https://github.com/CodinGame/monaco-vscode-api/blob/61752f5a8e77b1359c8648f8455d048d483c7e79/demo/src/setup.common.ts#L245-L247

kaisalmen: Generally a good idea, but not for this specific problem, because it is just a workaround. I want to introduce the view service here (derived from your demo) soon. I started that work on branch https://github.com/TypeFox/monaco-languageclient/tree/oct-vscode-web and may come back to that idea then. WDYT?

CGNonofr: Why not for this pecific problem? at least you have the guarantee to not impact anything outside the current session

kaisalmen: Ok, that became a bigger change, but hopefully for good. I moved the `userConfiguration` out of the editor app to the renamed `vscodeApiConfig`. If semanticHighlighting is now configured in the editorOptions this is now set to workspaceConfig.configurationDefaults.

Can you tell me why this is a such? https://github.com/TypeFox/monaco-languageclient/blob/env-views-options/packages/wrapper/test/wrapper.test.ts#L244-L245

It works as expected when I check the Langium Classic example, but I didn't expect `configuredByThem` as return in the test. I thought it is just true.

CGNonofr: I don't know :thinking: maybe you check it too soon?

kaisalmen: @CGNonofr a timeout works. Now the test ensure a worker factory is available, too. That prevented prevents some errors. The test didn't fail, but there where unexpected errors.

kaisalmen:
@CGNonofr this is now ready.

packages/wrapper/test/wrapper.test.ts
```diff
@@ -230,17 +230,24 @@ describe('Test MonacoEditorLanguageClientWrapper', () => {
...
 
-        // why is this configuredByTheme?
-        const semHigh = StandaloneServices.get(IConfigurationService).getValue('editor.semanticHighlighting.enabled');
-        expect(semHigh).toEqual('configuredByTheme');
+        // await wrapper.start();
+
+        const semHigh = await new Promise<unknown>(resolve => {
+            setTimeout(() => {
+                resolve(StandaloneServices.get(IConfigurationService).getValue('editor.semanticHighlighting.enabled'));
+            }, 1000);
```
CGNonofr: Are you sure you want to keep that timeout? it may make the test flacky

kaisalmen: I reduced it to 100ms already, but it only works with a brief timeout. Let's keep for now.

CGNonofr (approved):

PR #770, opened 2024-10-22
GitHub Pages for examples without backend

kaisalmen:
This PR makes the demos that don't require a backend available via GitHub Pages:
https://typefox.github.io/monaco-languageclient

This branch is based on the clangd branch, because I wanted to see if that works, but it does not, yet.

Problems:
- [x] [Statemachine example](https://typefox.github.io/monaco-languageclient/statemachine.html) does not work, because the launcher url (locale loading) is not treated properly.
- [x] [Clangd example](https://typefox.github.io/monaco-languageclient/clangd.html): The workers are broken after the build. Maybe they should be treated as assets. I don't know a way forward, yet. **Solution**: Manually copy the `clangd.js` and `clangd.worker.mjs` to the asset directory.
- [x] [Clangd example](https://typefox.github.io/monaco-languageclient/clangd.html): I used [mini-coi](https://github.com/WebReflection/mini-coi) to successfully get around SharedArrayBuffers being unavailable.
- [x] The based vsix github theme in [Langium Extended Example](https://typefox.github.io/monaco-languageclient/langium.html) is not loaded although the Plugin is configured. **Solution**: Don't let it be an asset.

@CGNonofr Any ideas how to solve the problems are appreciated.


CGNonofr:
> [Clangd example](https://typefox.github.io/monaco-languageclient/clangd.html): The workers are broken after the build. Maybe they should be treated as assets. I don't know a way forward, yet

I don't have any error in the console :thinking: 

> [Statemachine example](https://typefox.github.io/monaco-languageclient/statemachine.html) does not work, because the launcher url (locale loading) is not treated properly.

I'm not sure about the error you mention, the first errors I have are

image

Why the ts extension? was your worker transformed?

kaisalmen:
> I don't have any error in the console 🤔

You don't see those after pressing start?
image

> Why the ts extension? was your worker transformed?**

The uri of the file contains the ts extension. I need to change the way it is loaded.

kaisalmen:
@CGNonofr it seems to comes from this:
image

CGNonofr:
> You don't see those after pressing start?

Oh yes, but since there is a white square, I didn't go further

kaisalmen:
@CGNonofr clangd is working! 🎉

CGNonofr:
> @CGNonofr clangd is working! 🎉

YEAH, what was the issue?

kaisalmen:
> edit: Ho missed it! but why was it required? 🤔

Manually copy the clangd.js and clangd.worker.mjs to the asset directory.

> edit2: btw, downloading a 20mo wasm hurts a bit :D

Yes, it's big, but it contains all header files. Btw, you can navigate to them. 🙂 

> edit3: why is there a start button though? can't it start directly?

Yes, I may remove that again, It only makes sense if there is a dispose button as well. But on the other hand loading the big stuff only happens on button press, doesn't it? I have to check..

kaisalmen:
Files are synced via a MessageChannel using my [wtd-core](https://www.npmjs.com/package/wtd-core) library. This is still a prototype, but works smoothly already.

CGNonofr:
> Yes, it's big, but it contains all header files. Btw, you can navigate to them.

Isn't there any way to externalize them and download them only when needed?

CGNonofr:
> > Isn't there any way to externalize them and download them only when needed?
> 
> Haven't thought about that yet. Isn't the browser caching them?
> 
> GHP It is not yet mentioned in the README. Maybe we should not do this until we figured the usability things out.

It probably does depending on the response headers, but you probably only need only a few of them 99.9% of the time, that's a shame to download them all at first :)

kaisalmen:
Before pressing start:
image

After start:
image

kaisalmen:
@CGNonofr statemachine loading issue is also resolved.

kaisalmen:
@CGNonofr the last force push only added a sentence to the index page.

CGNonofr (approved):

PR #774, opened 2024-10-28
Add new example Application Playground

kaisalmen:
Use views service in new example.
@CGNonofr The purpose is not language client related (at least not yet). It should test the the views service integration and the [open collaboration tools](https://github.com/TypeFox/open-collaboration-tools) vscode web extension. ~~This is working within your demo, but not here, yet. I need to find what services are missing.~~
The extension is working, but after the connection is established to the server the are now errors not resulting from this example, but also broken in VSCode.dev and the monaco-vscode-api demo where it used to work before. I will open an issue int the OCT repo.

Navigation back to index page is added (for dev server and deployed version).


kaisalmen:
@CGNonofr this is ready for review. It will also evolve like the clangd example, but the views service is now working. The OCT extension issues need be resolved externally, first. This is not blocking this PR, IMO.

CGNonofr (approved):
LGTM

kaisalmen:
@CGNonofr the open collaboration extension is working, btw. The serving session crashed. So, it was my fault.

PR #777, opened 2024-10-31
Clangd example: Additional resources, volatile directory

kaisalmen:
With this change you can choose to load or not load the default workspace content.
In addition you can now put additional sources under `resources/clangd/workspace/volatile` and all cpp, c, h and hpp files in all sub dirs will be loaded.

monaco-vscode-api is updated to 10.1.4. This fixes the file navigation.


CGNonofr (approved):

kaisalmen:
@CGNonofr thank you. With this change you can fairly easily add larger amounts of header files for code navigation.

PR #808, opened 2024-12-17
Bugfix: Wrapper; Text model content is not properly updated with updateCodeResource

kaisalmen:
Text model content was not updated properly when updating code resources
Added additional unit tests
Variable naming alignment. Terminology is always modified and original derived from DiffEditor.


CGNonofr (approved):
LGTM

PR #818, opened 2025-01-06
App Playground; Show explorer view when search is included

### LINKED ISSUE #815: app playground: using search service make it the default tab
if using the search service then upon opening of the app playground, instead of seeing the explorer tab, it defaults to the search tab:
image

This is confusing. I couldn't find where to set it. It's more of a vscode issue than it is a monaco-languageclient issue.

Relevant wrapper config:
```
				serviceOverrides: {
					...getConfigurationServiceOverride(),
					...getKeybindingsServiceOverride(),
					...getLifecycleServiceOverride(),
					...getSearchServiceOverride(),
					...getExplorerServiceOverride()

				},

				viewsConfig: {
					viewServiceType: 'ViewsService',
					viewsInitFunc: defaultViewsInit
				},
```

kaisalmen:
Fixes #815

### DIFF
--- a/packages/examples/src/appPlayground/main.ts
+++ b/packages/examples/src/appPlayground/main.ts
@@ -4,7 +4,7 @@
  * ------------------------------------------------------------------------------------------ */
 
 import * as vscode from 'vscode';
-import { LogLevel } from 'vscode/services';
+import { getService, IWorkbenchLayoutService, LogLevel } from 'vscode/services';
 import { RegisterLocalProcessExtensionResult } from 'vscode/extensions';
 import getConfigurationServiceOverride from '@codingame/monaco-vscode-configuration-service-override';
 import getKeybindingsServiceOverride from '@codingame/monaco-vscode-keybindings-service-override';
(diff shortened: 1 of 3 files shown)


packages/examples/src/appPlayground/main.ts
```diff
@@ -128,6 +131,12 @@ export const runApplicationPlayground = async () => {
     const result = wrapper.getExtensionRegisterResult('mlc-app-playground') as RegisterLocalProcessExtensionResult;
     result.setAsDefaultApi();
 
+    // currently unused
+    await getService(IWorkbenchLayoutService);
+
+    // show explorer and not search
+    await vscode.commands.executeCommand('workbench.view.explorer');
```
CGNonofr, line 138: I'm not sure why it's required :thinking: it would prevent the storage service from being able to restore the last open view

kaisalmen: So, you are saying, it should not behave as it currently, but maybe it it does because the storage service is not used?

CGNonofr: I see no reason for the search service to be displayed by default :thinking:

The storage service not being used just prevents it from storing and restoring the active view, it shouldn't have any impact on the default view on a fresh new browser

kaisalmen: When the storage service is used. The positions are kept in local storage, but the search is always displayed, so there seems to be another issue:
image

CGNonofr: That weird :thinking:  you can try to play with the `defaultLayout.views` construct option :thinking: 

Maybe there is a bug or a race condition in the code

It probably has something to do with https://github.com/CodinGame/monaco-vscode-api/blob/c228c61f6dbcfebeeffa6769373ff356abacc0e3/src/service-override/views.ts#L427-L471, you can try debugging in there?

CGNonofr: `defaultLayout.views` allows to select which view should be displayed at first, I'm curious what happens it you force the explorer view

kaisalmen: Ok, it doesn't change the behaviour

CGNonofr: Ok so there is probably something that changes it later...

kaisalmen: Yes, that seems to be the case.

kaisalmen: Can you have a look at some point. I [suggested the workaround for now](https://github.com/TypeFox/monaco-languageclient/issues/815#issuecomment-2573328837).

kaisalmen:
This PR has become obsolete because of #823

Issue #819, opened 2025-01-06
app playground: listening for load progress


rubenfiszel:
Hi,

If you load https://typefox.github.io/monaco-languageclient/ghp_appPlayground.html, you should see that first it loads a white editor and then the theme is loaded and it becomes properly themed. That's content/color shift that I'd like to avoid for our users. It would likely be enough for us to wait for the themeService to be loaded and waiting for that to show a properly themed overlay/progress bar. I was wondering if there were any suggestions around getting:
- progress of items being loaded so we can bump the progress bar
- know when the themeService is sufficiently loaded that it would display the proper colors
- know when the full editor + workspace has been loaded in case we want to avoid all content shift.

CGNonofr:
In VSCode, the storage service is used to load the current theme faster (if it didn't change since the last visit)

So you can try:
- loading the storage service
- make sure the theme is set in the configuration BEFORE loading the services (by using initUserConfiguration)

kaisalmen:
A PR is up for adding the suggestion to the wrapper.

kaisalmen:
Issue review before new release. I will close this now. This is considered resolved.

PR #820, opened 2025-01-06
Ensure configuration is init before service init

kaisalmen:
The theme flickering is prevented by this when using the wrapper as suggested by @CGNonofr [here](https://github.com/TypeFox/monaco-languageclient/issues/819#issuecomment-2573197462).

### DIFF
--- a/packages/client/src/vscode/services.ts
+++ b/packages/client/src/vscode/services.ts
@@ -17,7 +17,7 @@ import { EnvironmentOverride } from 'vscode/workbench';
 import { Logger } from 'monaco-languageclient/tools';
 import { FakeWorker as Worker } from './fakeWorker.js';
 import { setUnexpectedErrorHandler } from 'vscode/monaco';
-import { getUserConfiguration, initUserConfiguration, updateUserConfiguration } from '@codingame/monaco-vscode-configuration-service-override';
+import { initUserConfiguration } from '@codingame/monaco-vscode-configuration-service-override';
 
 export interface MonacoEnvironmentEnhanced extends monaco.Environment {
     vscodeInitialising?: boolean;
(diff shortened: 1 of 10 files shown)


packages/client/src/vscode/services.ts
```diff
@@ -100,18 +100,24 @@ export const initServices = async (vscodeApiConfig: VscodeApiConfig, instruction
...
+            if (vscodeApiConfig.userConfiguration?.json !== undefined) {
+                await updateUserConfiguration(vscodeApiConfig.userConfiguration?.json);
+            }
         } else {
             envEnhanced.vscodeInitialising = true;
             instructions.logger?.debug(`Initializing vscode services. Caller: ${instructions.caller ?? 'unknown'}`);
 
+            if (vscodeApiConfig.userConfiguration?.json !== undefined) {
+                await initUserConfiguration(vscodeApiConfig.userConfiguration?.json);
+            }
```
CGNonofr, line 113: I guess we may want a way to initialize the default configuration without overriding it each time :thinking:

kaisalmen: Ah, yes, sorry, I make this mistake over and over. We have to merge what is already there.

kaisalmen: `getUserConfiguration` does not return (await hangs forever) before any service init is done. Is this expected? That call above is only done once, btw.

CGNonofr: Yes, it is expected

I'm not sure of the point of even merging the configuration :thinking:

kaisalmen: > I'm not sure of the point of even merging the configuration 🤔

Wasn't that the point of your first comment? "I guess we may want a way to initialize the default configuration without overriding it each time"  I am confused.

CGNonofr: I mean just calling initUserConfiguration when the service aren't initialized

I don't think I would want the configuration to be "restored" if I call initServices a second time :thinking:

kaisalmen: Ok, then the last commit matches what you expect.

kaisalmen: `vscodeApiInitialised` is stored in the "enhanced" global monaco environment and once services are initialized that code block is not hit again.

CGNonofr: I'm still concerned by the fast that you're updating the configuration when `vscodeApiInitialised` is true. I'm not sure when it's supposed to be called, but:
- I would change the file formatting
- It would change the property order
- It would reset user settings if the user changes it

I'm not sure about the point of it :thinking: if the developper want to keep the control of a settings, they can just use the defaultConfiguration in the workbench construct options. Initializing the user settings only exists so that the file is not empty, and to showcase what property the user may want to change

kaisalmen: Ok, then let's only do user config init before services are configured and only if set in the config. This is only done once. Even if you recall the `initServices`.
What the previous code did was wrong, because it always called `updateUserConfiguration`. I really should add a unit test for this actually.

kaisalmen: Ok, changes and unit test are up ⬇️

packages/examples/src/bare/client.ts
```diff
@@ -15,24 +15,24 @@ import { WebSocketMessageReader, WebSocketMessageWriter, toSocket } from 'vscode
...
 
 export const runClient = async () => {
     const logger = new ConsoleLogger(LogLevel.Debug);
     const htmlContainer = document.getElementById('monaco-editor-root')!;
     await initServices({
         serviceOverrides: {
             ...getConfigurationServiceOverride()
-        }
+        },
+        userConfiguration: {
```
CGNonofr, line 26: what about renaming it to `defaultUserConfiguration`?

kaisalmen: It is an API change and basically everything you pass is the default. I would keep it like this for now.

PR #839, opened 2025-02-04
clean subscriptions in react

zoran995:
Having the subscriptions stored in state will make it impossible to clean them as the `disposeOnTextChanged` callback will always access the old array without `onTextChangedSubscriptions` being in dep array


kaisalmen (approved):
Thank you for spotting this. LGTM

PR #842, opened 2025-02-05
Integrate python debugger

kaisalmen:
I chose to integrate the debug server directly into the container.
graalpy is now started properly. All workspace files are transported over to the container on start.

The debugger client and server are generic. On the server side the container ensure what DAP capable software is available. The start command for the debugger is send from the client on init. This way it is completely exchangable.
A launch config is supplied and the files are transferred to the debugger on start.

Fixes #690

PR #847, opened 2025-02-07
refactor: enhance code performance, maintainability and extendibility

ahmed-s-fatahallah:
replace the too many `useEffect` and `useCallbacks` with one `useEffect` to instantiate and update the editor and another effect to `dispose` the models' Subscriptions when the component unmounts.

### DIFF
--- a/.github/workflows/debugger-py.yml
+++ b/.github/workflows/debugger-py.yml
@@ -1,0 +1,63 @@
+name: Python Debugger Image
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - issue-690-debugger
+
+env:
+  REGISTRY: ghcr.io
+  REPO_NAME: ${{ github.repository }}
+  PATH_CONTEXT: ./packages/examples/resources/debugger
+  CONTAINER_NAME: debugger-py
+
+jobs:
+  image-debugger-py:
+    name: Build & Deploy Python Debugger
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+
+    timeout-minutes: 15
+    steps:
+    - name: Checkout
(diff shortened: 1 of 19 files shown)


kaisalmen:
Hi @ahmed-s-fatahallah thank for this PR. It made me realize the subscriptions for the onDidChangeContent should not be in the react component at all. I opened #849 to correct that.

Please fix the linting problems, then we can review the changes.

ahmed-s-fatahallah:
> Hi @ahmed-s-fatahallah thank for this PR. It made me realize the subscriptions for the onDidChangeContent should not be in the react component at all. I opened #849 to correct that.
> 
> Please fix the linting problems, then we can review the changes.

You are welcome, I will fix the linting problems as soon as possible.

Looking forward for your review

kaisalmen:
@ahmed-s-fatahallah please rebase your PR. #849 removes the need to deal with subscriptions on the react level, therefore you need to change less code. Sorry for the inconvenience. I will review the changes after the rebase.

ahmed-s-fatahallah:
@kaisalmen Hello Kai,
It looks like the onTextChanged prop on the MonacoEditorReactComp isn't receiving the textChanges parameter after the latest update. Should it capture the changes in the code editor? If yes, it seems to be broken; if no, can you clarify its purpose?

Thanks

kaisalmen:
> It looks like the onTextChanged prop on the MonacoEditorReactComp isn't receiving the textChanges parameter after the latest update.

Hi @ahmed-s-fatahallah we already had unit tests that verify onTestChanges and they are still green. Why do you come to that conclusion? I also test it manually. There is one thing though: If you use the views service the callbacks (appPlayground and python example (after merge of #842)) those callbacks don't get registered because "vscode" creates the editors depending which resources/uri is opened.

PR #849, opened 2025-02-08
Move text changes handling from react component to the wrapper

kaisalmen:
This was a bad design choice that is corrected. The react component should be thinnest possible layer around the wrapper.


kaisalmen:
@CGNonofr this is now ready for review. I added a unit test.

CGNonofr (approved):
LGTM

kaisalmen:
@CGNonofr thank you. So I guess you agree/confirm that lesser logic in react is a plus.

CGNonofr:
> @CGNonofr thank you. So I guess you agree/confirm that lesser logic in react is a plus.

Sure! I don't want code to be duplicated between framework implementations

PR #859, opened 2025-02-14
Handle languageclient errors cases more robustly

kaisalmen:
I moved some tests around (wrong places) as well.
Fixing the wrapper handling in case of error introduced required changes to the react component and to several tests.

After thinking about #855

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11378,7 +11378,6 @@
       "license": "MIT",
       "dependencies": {
         "@codingame/monaco-vscode-api": "~14.0.2",
-        "@codingame/monaco-vscode-configuration-service-override": "~14.0.2",
         "@codingame/monaco-vscode-editor-api": "~14.0.2",
         "@codingame/monaco-vscode-editor-service-override": "~14.0.2",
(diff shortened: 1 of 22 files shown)


kaisalmen:
@CGNonofr I finally understood why awaiting the language client on connection error broke the await in the test and it is now fixed with: [8c2b8f8](https://github.com/TypeFox/monaco-languageclient/pull/859/commits/8c2b8f8a64b14640041b09c47dfe2b3affe2a2c0)

In a real application you should always define your own error handler!

kaisalmen:
@CGNonofr last commit is for the problem @cdietrich discovered with the react component introduced by 6.3.0 version. We now have 78 unit tests.

CGNonofr (approved):
not easy to review! :cry:  LGTM though

kaisalmen:
@CGNonofr sorry, one I started fixing the tests the domino effect started.

kaisalmen:
@CGNonofr These changes made the `Unknown Error: ResizeObserver loop completed with undelivered notifications.` in the tests go away. The question now is why exactly, but if we find an answer this could be interesting. The `monaco-vscode-api` version was unchanged, btw.

PR #860, opened 2025-02-14
refactor: replace all `useEffect` and `useCallback` with one `useEffect`

ahmed-s-fatahallah:
This pull request refactors the index.tsx file in the wrapper-react package to simplify and improve the usage of  `useEffect` and `useCallback` hooks. 

### Changes Made:
Replaced multiple `useEffect` and `useCallback` hooks with a single `useEffect`.

Enhanced container reference initialization and cleanup logic.

Simplified error handling for initMonaco, startMonaco, and disposal processes.

Impact:
These changes reduce code complexity and enhance maintainability, making it easier to manage the lifecycle of the Monaco editor component.

Please review the changes and provide any feedback or suggestions. Thank you!

Also, I noticed that you always assign the `containerRef.current` to `wrapperConfig.htmlContainer` even if the wrapperConfig has an htmContainer key and value which always overrides the user-specified container so I thought about assigning the `ref` to the `wrapperConfig.htmlCotnainer` only if it doesn't exist to give the user the freedom to specify any container he wants but when I did so the container specified inside the `wrapperConfig` won't have any fixed height which will make the editor keep growing indefinitely, so I am thinking about always adding the editor inside that div inside the react component and if the user specified a container inside the `wrapperConfig` then that container will be used as a container to this div.

so I was asking if I should make this change and create a new PR for it?

Thanks

### DIFF
--- a/packages/wrapper-react/src/index.tsx
+++ b/packages/wrapper-react/src/index.tsx
@@ -57,38 +57,30 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
     }, [wrapperConfig]);
 
     const initMonaco = useCallback(async () => {
-        if (!wrapperRef.current.isInitializing()) {
-            if (containerRef.current) {
-                wrapperConfig.htmlContainer = containerRef.current;
-                await wrapperRef.current.init(wrapperConfig);
-            } else {
-                throw new Error('No htmlContainer found! Aborting...');
-            }
+        if (containerRef.current) {
+            wrapperConfig.htmlContainer = containerRef.current;
+            await wrapperRef.current.init(wrapperConfig);
         } else {
-            await wrapperRef.current.getInitializingAwait();
+            throw new Error('No htmlContainer found! Aborting...');
         }
     }, [wrapperConfig]);
 
     const startMonaco = useCallback(async () => {
-        if (!wrapperRef.current.isStarting()) {
-            if (containerRef.current) {
-                try {
-                    wrapperRef.current.registerTextChangeCallback(onTextChanged);
-                    await wrapperRef.current.start();
-                    onLoad?.(wrapperRef.current);
-                    handleOnTextChanged();
-                } catch (e) {
-                    if (onError) {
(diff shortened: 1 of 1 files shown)


kaisalmen:
Hi @ahmed-s-fatahallah thank you. I generally agree with your approach. I fixed some problems with the wrapper today that lead to changes in the react component which lead to conflicts again. Sorry for that.

But, I basically agree with your idea. The main `useEffect` could be structured as such: 
```js
    useEffect(() => {
        (async () => {
            await destroyMonaco();
            await initMonaco();
            await startMonaco();
        })();
    }, [wrapperConfig]);
```
With `await` the code is cleaner and therefore I propose a self-invoking function in the use effect. Your code blocks could just be moved into those three async functions.

Regarding the `htmlContainer` the react component should specify the container. If `wrapperConfig.htmlContainer` is defined a warning should be logged. You can use `wrapper.getLogger()` for that.

kaisalmen:
@ahmed-s-fatahallah you could have just rebased your branch to the latest main. No need to close the PR.

ahmed-s-fatahallah:
@kaisalmen I synced your changes to my fork and it closed the PR automatically, sorry for that, I will create another PR today with the changes. 
BTW, I also wanted to talk about your proposal for the main `useEffect` doing it like 
```javascript
    useEffect(() => {
        (async () => {
            await destroyMonaco();
            await initMonaco();
            await startMonaco();
        })();
    }, [wrapperConfig]);
```
will require at least 3 `useCallback` hooks to persist these 3 functions which I think we can remove them and just declear these functions inside only 1 `useEffect` because after all the more `useCallback` hooks the more dependencies arrays react needs to keep track of and if we this dependency array is gonna be extended in the future then it will need to be changed in 4 places so I think we can get rid of all these `useCallback` hooks and just declear these functions inside the main `useEffect`
    
 Also, about the container, if the react component will always define the container why is there an option to add an htmlContainer in the configWrapper obj shouldn't it be removed, and the container is passed as a second parameter to the `init` function ?

PR #862, opened 2025-02-15
refactor: replace all `useEffect` and `useCallback` with one `useEffect`

ahmed-s-fatahallah:
#### Changes Made:
- Replaced multiple `useEffect` and `useCallback` hooks with a single, cohesive `useEffect`.
- Improved container reference management and initialization.
- Streamlined error handling and disposal processes for better reliability.

#### Impact:
These adjustments are intended to reduce code complexity, making the lifecycle management of the Monaco editor component more straightforward and maintainable.

Please review these changes and provide any feedback. Thank you!

### DIFF
--- a/packages/wrapper-react/src/index.tsx
+++ b/packages/wrapper-react/src/index.tsx
@@ -3,7 +3,7 @@
 * Licensed under the MIT License. See LICENSE in the package root for license information.
 * ------------------------------------------------------------------------------------------ */
 
-import React, { type CSSProperties, useCallback, useEffect, useRef } from 'react';
+import React, { type CSSProperties, useEffect, useRef } from 'react';
 import { MonacoEditorLanguageClientWrapper, type TextContents, type WrapperConfig } from 'monaco-editor-wrapper';
 
 export type MonacoEditorProps = {
@@ -12,8 +12,7 @@ export type MonacoEditorProps = {
     wrapperConfig: WrapperConfig,
     onTextChanged?: (textChanges: TextContents) => void;
     onLoad?: (wrapper: MonacoEditorLanguageClientWrapper) => void;
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    onError?: (e: any) => void;
+    onError?: (e: unknown) => void;
 }
 
 export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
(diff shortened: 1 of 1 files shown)


ahmed-s-fatahallah:
Thank you for your suggestion to group each step into separate functions and call them inside an async IIFE. As mentioned in my last comment, I believe that declaring these functions outside the main `useEffect` might necessitate wrapping them in `useCallback`, which I think we can avoid.

I appreciate your patience and help!

packages/wrapper-react/src/index.tsx
```diff
@@ -30,72 +29,49 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
     const containerRef = useRef<HTMLDivElement>(null);
 
     useEffect(() => {
```
kaisalmen, line 31: We need a second `useEffect` for destruction, otherwise the editor resources are not properly destroyed:
```js
useEffect(() => {
  return () => {
    destroyMonaco();
  };
}, []);
```

packages/wrapper-react/src/index.tsx
```diff
@@ -30,72 +29,49 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
-        return () => {
-            destroyMonaco();
+        const destroyMonaco = async () => {
```
kaisalmen, line 32: You can move all methods outside the use effects, so you can use them in either one.

packages/wrapper-react/src/index.tsx
```diff
@@ -30,72 +29,49 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
-
-    const initMonaco = useCallback(async () => {
-        if (containerRef.current) {
-            wrapperConfig.htmlContainer = containerRef.current;
-            await wrapperRef.current.init(wrapperConfig);
-        } else {
-            throw new Error('No htmlContainer found! Aborting...');
-        }
-    }, [wrapperConfig]);
+        const initMonaco = async () => {
```
kaisalmen, line 41: You can move all methods outside the use effects, so you can use them in either one.

packages/wrapper-react/src/index.tsx
```diff
@@ -30,72 +29,49 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
-                wrapperRef.current.registerTextChangeCallback(onTextChanged);
-                await wrapperRef.current.start();
-                onLoad?.(wrapperRef.current);
-                handleOnTextChanged();
-            } catch (e) {
-                if (onError) {
-                    onError(e);
-                } else {
-                    throw e;
+        const startMonaco = async () => {
```
kaisalmen, line 50: You can move all methods outside the use effects, so you can use them in either one.

packages/wrapper-react/src/index.tsx
```diff
@@ -30,72 +29,49 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
-            throw new Error('No htmlContainer found! Aborting...');
-        }
-    }, [onError, onLoad, onTextChanged]);
+        };
 
-    const handleOnTextChanged = useCallback(() => {
-        if (!onTextChanged) return;
-    }, [onTextChanged, wrapperConfig]);
+        (async () => {
+            await destroyMonaco();
```
kaisalmen, line 69: I know this was like this before, but it makes no sense, because if you have the second use effect for destruction, the init of the wrapper ensures everything was disposed.

kaisalmen (changes requested):
We are almost there, we need to properly handle destruction. All my comments are basically related to this.
I tested your changes and the dispose is no longer working properly.

ahmed-s-fatahallah:
@kaisalmen First of all thanks for your comments.
However, wouldn't it be simpler to move the `destroyMonaco` invocation inside the main `useEffect` cleanup function, rather than creating a new `useEffect` with an empty dependencies array just for it?

Also, moving all methods outside the main `useEffect` will require wrapping them inside a `useCallback` which I think we can avoid by just including the methods inside the main `useEffect`.

Lastly, could you please share how you tested the `destroyMonaco` function? I want to ensure I test it before pushing, as all unit tests were passing with my last push.

kaisalmen:
> Also, moving all methods outside the main useEffect will require wrapping them inside a useCallback which I think we can avoid by just including the methods inside the main useEffect.

@ahmed-s-fatahallah it is this example:
https://typefox.github.io/monaco-languageclient/ghp_react_statemachine.html
or locally:
https://github.com/TypeFox/monaco-languageclient/blob/main/packages/examples/react_statemachine.html

packages/wrapper-react/src/index.tsx
kaisalmen: > Also, moving all methods outside the main useEffect will require wrapping them inside a useCallback which I think we can avoid by just including the methods inside the main useEffect.

Yes, thank you, that is the cleaner solution and I just verified that it works. It would be like that with only one `useEffect` 
```ts
    useEffect(() => {
       ...
        (async () => {
            await initMonaco();
            await startMonaco();
        })();
        return () => { destroyMonaco(); };
    }, [wrapperConfig]);
```

packages/wrapper-react/src/index.tsx
kaisalmen: See above ⬆️

packages/wrapper-react/src/index.tsx
kaisalmen: See above ⬆️

packages/wrapper-react/src/index.tsx
kaisalmen: See above ⬆️

packages/wrapper-react/src/index.tsx
kaisalmen: The above code snippet (second reply to first comment) takes this into account already.

kaisalmen:
@ahmed-s-fatahallah I have updated all comments and if you use the clean-up function for the `useEffect` like you suggested this PR is ready to be merged. 🎉

ahmed-s-fatahallah:
> @ahmed-s-fatahallah I have updated all comments and if you use the clean-up function for the `useEffect` like you suggested this PR is ready to be merged. 🎉

@kaisalmen 
I have pushed the clean-up function change

Thanks a lot for your patience and support, and thanks a lot for this amazing repo 🙏

kaisalmen (approved):
LGTM! Thank you @ahmed-s-fatahallah and sorry for the bumpy ride due to the rebases.

PR #872, opened 2025-02-19
implement async task queue to synchronize react wrapper initialization useEffect

### LINKED ISSUE #871: React strict mode race condition
When running react-wrapper it will run into race condition during initialization and cleanup and throw an error 
```js
Uncaught (in promise) Error: file 'extension-file://typefox.statemachine-example/extension/react-statemachine-configuration.json/' already exists
```

Steps to reproduce:
- Start `React: Langium Statemachine Language Client & Language Server (Worker)` example (only reproducible in local dev mode)
- Check strict mode
- Start the editor
- Check the error in the console (when using webpack error overlay, it will be annoying with popup)

When running in strict mode react will run useEffect twice but won't wait for async code to complete before running the cleanup, so the execution will basically be 
`initializing 1` -> `destroying 1` -> `initializing 2` -> `destroyed 1` -> `initialized 1` -> `starting 1` -> `started 1`
the error is thrown inside initializing for second time. 
Because both effect and cleanup are asynchronous, one possible solution would be to implement a basic task queue to ensure that nothing runs in parallel.

zoran995:
Fixes #871 

This will ensure that dispose is not invoked before the editor is fully started. This will slow down initialisation in development mode, as the editor will be fully started, then destroyed and started again; however, in production, it should work the same as in the main (unless some unexpected re-renders occur on the client side).

### DIFF
--- a/packages/wrapper-react/src/index.tsx
+++ b/packages/wrapper-react/src/index.tsx
@@ -5,6 +5,7 @@
 
 import React, { type CSSProperties, useEffect, useRef } from 'react';
 import { MonacoEditorLanguageClientWrapper, type TextContents, type WrapperConfig } from 'monaco-editor-wrapper';
+import { TaskQueue } from './task-queue.js';
 
 export type MonacoEditorProps = {
     style?: CSSProperties;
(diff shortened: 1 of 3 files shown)


zoran995:
This potentially uncovered different issue as it doesn't completely initialize the second time and it hangs at starting the language clients still looking into this

PR #873, opened 2025-02-20
fix: old monaco editor instance is not destroyed before a new one init

ahmed-s-fatahallah:
### Changes Made:

Added a call to `destroyMonaco()` before initializing a new instance with `initMonaco()` and `await` it.

Added `if (wrapperRef.current.isStopping() === false)` to make sure the editor is not stopping to prevent double editors in react strict mode because in strict mode react will execute `useEffect` hooks twice

This ensures that the previous instance is completely disposed of before starting a new one, improving performance and stability.

Removed the cleanup function from the main `useEffect` because it runs each time the wrapper config changes

Added a new `useEffect` with an empty dependencies array and a cleanup function to make sure that the Monaco instance is disposed before the component unmounts

### DIFF
--- a/packages/wrapper-react/src/index.tsx
+++ b/packages/wrapper-react/src/index.tsx
@@ -66,15 +66,25 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
         };
 
         (async () => {
-            await initMonaco();
-            await startMonaco();
+            if (wrapperRef.current.isStopping() === false) {
+                await destroyMonaco();
+                await initMonaco();
+                await startMonaco();
+            }
         })();
 
-        return () => {
-            destroyMonaco();
-        };
+    }, [wrapperConfig, onTextChanged, onLoad, onError]);
 
-    }, [wrapperConfig]);
+    useEffect(()=>{
+        return ()=>{
+            try {
+                wrapperRef.current.dispose();
+            } catch {
+                // The language client may throw an error during disposal.
+                // This should not prevent us from continue working.
+            }
+        };
+    },[]);
 
     return (
         <div


cdietrich:
unfortunately this does not work for me
https://github.com/TypeFox/monaco-languageclient/discussions/858#discussioncomment-12260931

zoran995:
Note: this doesn't resolve the error thrown in react strict mode

ahmed-s-fatahallah:
> Note: this doesn't resolve the error thrown in react strict mode

what error are you talking about because I tested it in strictmode and no errors were thrown.

zoran995:
It works now after the latest changes, thanks. It might be good to have a unit test to verify this and avoid regressions in the future.

Although not introduced in this PR, should we also include `onError`, `onLoad`, and `onTextChanged` in the effect dependency array to avoid calling old function

ahmed-s-fatahallah:
> It works now after the latest changes, thanks. It might be good to have a unit test to verify this and avoid regressions in the future.
> 
> Although not introduced in this PR, should we also include `onError`, `onLoad`, and `onTextChanged` in the effect dependency array to avoid calling old function

Agree with adding unit tests for this part as the current unit tests were passing while we had this bug so I think the current tests don't cover it.

Also, I have added `onTextChanged`, `onLoad`, and `onError` to the dependencies array to make sure we are not using an old function as you have suggested 🙏

cdietrich:
we need for @kaisalmen to come back from vacation

kaisalmen:
@ahmed-s-fatahallah thank you. I am back and catching up right now...

kaisalmen (approved):
Thank you. Let's merge it as such and add a unit test later. I would like to create a pre-release tomorrow.

ahmed-s-fatahallah:
> @ahmed-s-fatahallah thank you. I am back and catching up right now...

Welcome back, I hope you had a nice vacation. 😊

kaisalmen:
> Welcome back, I hope you had a nice vacation. 😊

Thank you, yes I had a good time.

PR #879, opened 2025-02-22
sync files with language server

justinmann:
* Created a custom FileSystemProvider to capture file changed
* Send file change/delete via new websocket message on the existing channel
* Use a /tmp folder path so the server can safely persist the files to disk
* Use file update time to make sure out-of-order websocket message does not overwrite a valid version
* Add getWorkingCopyServiceOverride so it autoSave setting would work
* defaultHtmlAugmentationInstructions was adding a div with no class that was screwing up the layout which is why your CSS needed to use "95vh". This approach makes it scales to 100% of parent which is much easier to use that viewport fixed percentage.
* Sends a "pipInstall" command that create a venv on the servers and installs from requirements.txt

TODO:
* The custom websocket message needs to be filtered out and not forwarded to Pyright not sure how to do this
* Debugger needs work, I had to disable it because it seems to have its file syncing but one-time
* If you add a new file, then pyright does not detect it and will fail to import. I tried to add --verbose to pyright to debug this, but I cannot see any output. "createServerProcess" seems to be messing around with stdout, so I have no idea where the logs for pyright are going.
* Get imports from external libraries to work, according to the import resolution for PyRight it seems to assume a single project is running. I have no idea how to convince it use the venv for each active workspace.
https://github.com/microsoft/pyright/blob/main/docs/import-resolution.md


kaisalmen:
Hi @justinmann thank you for opening this PR. Is this ready for review? It has some TODOs. Please use "Convert to draft" if still WIP.
I saw you want to introduce prettier. We can discuss this, but please do not enforce 80 characters and trailing commas.

Let us know when this is ready for review. Thanks.

justinmann:
This is not ready for review. It is more of a proof of concept. Without fixing the last two issues under TODO, it is sort of pointless. Those issues seem be related how Pyright functions, and I have no idea how to debug them. 

I would love to discuss this:
https://www.linkedin.com/in/thejustinmann/

kaisalmen:
@justinmann I am closing this PR due to inactivity. If you want to continue this work, please rebase your changes and we could re-open the PR.

PR #889, opened 2025-03-03
Update dependencies, fix tests, language clients config changes

kaisalmen:
Contains:

- Wrapper changes:
  - language client behavior for init, start and dispose is now driven by `WrapperConfig` and can be optionally configured differently from the default behavior
  - worker loader tests: now use Promises instead of fixed timeouts
  - init, start, dispose await previous executions of themselves
- react comp:
  - Update re-render test
- dependency updates
- Update next versions for testing
- Updated CHANGELOGs


kaisalmen:
@CGNonofr because there were a couple of issues with the react wrapper I want to release a preview version this time and await feedback. I added some improvements that should improve robustness. The language client behavior configuration makes it behave the same in pure wrapper or react component context.

kaisalmen:
@CGNonofr @cdietrich and @ahmed-s-fatahallah new preview versions are available that are based on the [last commit ](https://github.com/TypeFox/monaco-languageclient/pull/889/commits/80eae381d6abcb2562164830bc59ca6adbf86889);
monaco-languageclient@9.4.1-next.0
monaco-editor-wrapper@6.5.0-next.0
@typefox/monaco-editor-react@6.5.0-next.0

cdietrich:
@kaisalmen will check tomorrow morning

packages/wrapper-react/src/index.tsx
```diff
@@ -28,62 +28,56 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
-        return ()=>{
-            try {
-                wrapperRef.current.dispose();
-            } catch {
-                // The language client may throw an error during disposal.
-                // This should not prevent us from continue working.
-            }
-        };
+        (async () => {
+            await destroyMonaco();
```
ahmed-s-fatahallah, line 79: We need to call `detroyMonaco()` inside the cleanup function that is returned from this `useEffect` to call it before the component unmounts not inside the `useEffect` call back as the `useEffect` main callback with the empty dependencies array is called after the component is mounted for the first time only

you can check the `React: Langium Statemachine Language Client & Language Server (Worker)` example the editor instance is not destroyed when you click the `dispose` button

it should be like this
```typescript
useEffect(()=>{
        return ()=>{
            (async () => {
                await destroyMonaco();
            })();
        }
    },[]);
```

packages/wrapper-react/src/index.tsx
```diff
@@ -28,62 +28,56 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
-            } else {
-                throw new Error('No htmlContainer found! Aborting...');
-            }
-        };
+    const initMonaco = async () => {
+        if (containerRef.current) {
+            wrapperConfig.htmlContainer = containerRef.current;
+            await wrapperRef.current.init(wrapperConfig);
+        } else {
+            throw new Error('No htmlContainer found! Aborting...');
```
ahmed-s-fatahallah, line 45: Getting these functions declarations outside the `useEffect` won't introduce bugs but it will cause them to be redeclared every time this component is rendered it is not a big deal and the gc will collect the old functions but after the component initially mounts the `useEffect` will have a reference to the first function declarations and when the component rerenders these functions will be redeclared with a new reference but the `useEffect` would still hold on the old functions references because of closure so basically we will have 6 functions in memory 3 of them are useless, still it won't have that much of an impact on memory but I think it would be nice to have this optimization, am I right?

so I think here we have 2 solutions :

1. Wrap these 3 functions inside 3 `useCallback` hooks and give them the same dependencies array as the `useEffect`. But tbh, I don't like this solution because of maintainability issues. You will always have to make sure that all these hooks have the same dependencies arrays to avoid introducing unintentional bugs. Also, `useCallback` has other drawbacks.
2. Get these functions declarations inside the `useEffect` again to make sure that they are redeclared only when the `useEffect` is rerun and there is no reference to the old ones so the gc can collect them right away and for the `destroyMonaco()` inside the other `useEffect` it is just one line inside the function we can just call `wrapperRef.current.dispose();` inside the other `useEffect` cleanup function, and tbh I prefer this approach more because it is cleaner and less code and better for maintainability.

packages/examples/src/langium/statemachine/config/wrapperStatemachineConfig.ts
```diff
@@ -37,18 +37,23 @@ export const createLangiumGlobalConfig = async (params: {
...
-        statemachine: {
-            clientOptions: {
-                documentSelector: ['statemachine']
-            },
-            connection: {
-                options: {
-                    $type: 'WorkerDirect',
-                    worker: params.worker,
-                    messagePort: params.messagePort,
+    const languageClientConfigs: LanguageClientConfigs | undefined = params.useLanguageClient && params.worker ? {
```
ahmed-s-fatahallah, line 40: I, also, noticed here that the `createLangiumGlobalConfig` is an async function and you are awaiting it when you call it although there is nothing async inside of it. Am I right or am I missing something ?

ahmed-s-fatahallah (commented):
Awesome work and effort @kaisalmen, and thanks alot for asking for my review

packages/wrapper-react/src/index.tsx
kaisalmen: Thank you, I didn't catch that. I can see the difference locally.

packages/wrapper-react/src/index.tsx
kaisalmen: Ok, let's go with option 2 then. I will copy the `disposeMonaco` function  (renamed from `destroyMonaco`) to the second `useEffect` and add a comment that is the exact same function and only copied to prevent declaring various `useCallback`.

packages/examples/src/langium/statemachine/config/wrapperStatemachineConfig.ts
kaisalmen: Thanks, that is probably a left-over from a previous version. I will correct that.

packages/wrapper-react/src/index.tsx
kaisalmen: Done. See next commit ⬇️

packages/wrapper-react/src/index.tsx
kaisalmen: Done. See next commit ⬇️

packages/examples/src/langium/statemachine/config/wrapperStatemachineConfig.ts
kaisalmen: Done. See next commit ⬇️

kaisalmen:
@ahmed-s-fatahallah and @cdietrich I have made the automatically flags optional  and introduced a fourth one for handling the worker dispose. Apart from comment implementations I also aligned the naming a bit further (it is always called dispose now, not destroy or stop).

One thing that I see and don't yet understand in the react statemachine example: After disposing the editor and language client it does not work again when restarting. In the statemachine example using the wrapper directly this works fine. I need to debug it and find out why.

kaisalmen:
@CGNonofr this is now ready for final review. The problem regarding the language server workers is gone. It was disposed although it shouldn't have. I aligned the naming and usage.

@ahmed-s-fatahallah the need for `if (!wrapperRef.current.isDisposing()) {` is gone in the component as the main `dispose` function in the wrapper properly handles the situation.

I will push another update incrementing the next versions once more.

kaisalmen:
monaco-languageclient@9.4.1-next.1
monaco-editor-wrapper@6.5.0-next.1
@typefox/monaco-editor-react@6.5.0-next.1

are available

ahmed-s-fatahallah:
It works fine for me too 🙏

CGNonofr:
> @CGNonofr if you have no objections we could merge it.

Go ahead!

kaisalmen:
> Go ahead!
👍 @CGNonofr can you give the formal approval

CGNonofr (approved):

cdietrich:
hmmm in one of my examples i see this logged
Initialization of vscode services can only performed once!

kaisalmen:
@cdietrich between next.1 and final release there were no further code changes

kaisalmen:
> also the language server does not get disposed. need to check

@cdietrich you can configure this via the wrapper:
https://github.com/TypeFox/monaco-languageclient/blob/main/packages/wrapper/src/wrapper.ts#L22-L28

`automaticallyDisposeWorkers` must be set to "true" if you want to dispose the LS worker on dispose of the client. Reason behind is that it usually not needed to throw it away every time.

> Initialization of vscode services can only performed once!

This is not new. It is only a debug message. If you re-init, it will always be there after first init call.

kaisalmen:
I will add comments to the automatically... flags to highlight the default.

cdietrich:
Problem is langium based worker. It keeps old documents

cdietrich:
I this case I end up with two workers

kaisalmen:
ok, then `automaticallyDisposeWorkers` should do the trick for you. It should not bring up another worker if it already exists. 🤔

cdietrich:
Am not sure if my react usage is correct.

cdietrich:
automatically dispose works seems to work.
but i still see the old docs in the worker ....

cdietrich:
indeed. there server has two documents after
['file:///workspace/model31.hello', 'file:///workspace/model28.hello']
=> i wonder if these also need to be disposed client side

cdietrich:
indeed the client sends two didOpen

kaisalmen:
So, you need to perform some closure operations on the application level. Do we need general callbacks for that?

cdietrich:
hmmmm. the problem is how to get access to the language client otherwise.

cdietrich:
this might even be a race. when i debug here:
grafik
it seems to work .....

cdietrich:
but shouldnt the docs have been purged before
grafik

kaisalmen:
@cdietrich my understanding is that the documents will be sent as long as they are open in the editor and when a new language client is initialized. How did you open those two documents?
Those documents need to be closed, then they should not be re-sent once a new language client starts.

Do you miss a onDiposed callback that is invoked after disposal (similar to the onLoad callback)?

cdietrich:
@kaisalmen did you check out my branch
i just do a 

initially load react component
on button click update the edited model

cdietrich:
https://github.com/cdietrich/monaco-editor-react-new

kaisalmen:
@cdietrich provide a fixed uri as such (you are creating a new model with a button press):

```ts
modified: {
  uri: model.uri,
  text: model.content
```

Also, the button re-inits everything. Just use use a wrapper reference:
```tsx
 let wrapper: MonacoEditorLanguageClientWrapper;
 const clickMe = () => {
    wrapper.updateCodeResources({
      modified: {
        uri: "/workspace/demo.hello",
        text: "person B Hello B! Hello Person2!"
      }
    });

  }
  return (
    <div style={{ 'height': height }} >
      <button onClick={clickMe}>Click Me</button>
      <MonacoEditorReactComp
      wrapperConfig={wrapperConfig}
      onLoad={(w) => {
        wrapper = w;
      }}
    style={{ 'height': '100%' }}

        />
    </div>
```

cdietrich:
@kaisalmen 

hmmmm on a first try it did not work

to test: after button klick

grafik
a should not be there

the problem is in our real scenario the uris are calculated external and different on the before and after button click.
so using the same uri wont work

cdietrich:
hmmm maybe i also get the modified thing wrong? does this relate to the diff editor?

cdietrich:
its initially called with
{text: 'person A Hello A! Hello Person1!', fileExt: 'hello', enforceLanguageId: 'hello'}
=> no uri
thats my bad

cdietrich:
need to debug the different uri case

kaisalmen:
@cdietrich you can change the uri of course. It can be something like `/workspace/tester.hello` in the intial case:
```ts
modified: {
  uri: model.uri,
  text: model.content
}
```
instead of:
```ts
modified: {
  text: model.content,
  fileExt,
  enforceLanguageId: model.languageId
}
```
 modified is for the editor or writable part of a diff editor, original is or the diff editor (the read-only part), but not needed in your case.

If you want to get rid of old models you have to dispose them (iterate over models via monaco-editor api)

cdietrich:
hmmm
it should dispose it here

updateEditorModels(modelRefs: ModelRefs) {
        let updateMain = false;
        let updateOriginal = false;

        if (modelRefs.modelRefModified) {
            this.modelRefModified?.dispose();
            this.modelRefModified = modelRefs.modelRefModified;
            updateMain = true;
        }
        if (modelRefs.modelRefOriginal) {
            this.modelRefOriginal?.dispose();
            this.modelRefOriginal = modelRefs.modelRefOriginal;
            updateOriginal = true;
        }
        
        bit i feel it does not properly await.
        thus the races i see

kaisalmen:
Do it without react first. Get the logic straight. Everything monaco-editor or vscode api related things you have to do in TS/JS anyway.

Can we continue this in an issue on your repo. This is unrelated to the PR.

kaisalmen:
@cdietrich the python example has a regular and a react client and they share most of the logic, still the react one was created only recently. Testing things in TS and adopting it to react later is easier, IMO.

PR #891, opened 2025-03-07
Model handling improvements

kaisalmen:
- Old model references are disposed after editor model was updated.
- uri is now required to be specified in any case (code resources update or model update). We were creating random uris that become problematic when dealing with language servers. This change enforces users to "think" about uris in the context of their application.
-  Fix for #892: Dispose on init is done automatically, but you can now configure it to not do it. New unit test shows this.

Thanks to @cdietrich for finding this problem.


kaisalmen:
@cdietrich and @CGNonofr the tests became unstable due to the model disposal now properly happening. I had to defer some actions via `setTimeout` otherwise cancellation errors where thrown deep inside monaco.

I will emit pre-releases 6.6.0-next.0 in a minute...

kaisalmen:
Btw, the problem in the tests did not occur in the browser. So it can be a vitest problem to, but also the timings in the browser are different (human interaction).

cdietrich:
@kaisalmen is it ok if i move testing to monday morning?

kaisalmen:
> is it ok if i move testing to monday morning?

Of course, my response times will likely be bad Monday and Tuesday, because we gather in Kiel for two days.

cdietrich:
> Of course, my response times will likely be bad Monday and Tuesday, because we gather in Kiel for two days.

have fun with the team

kaisalmen:
@CGNonofr FYI these where the problems I needed to circumvent (taken from https://github.com/TypeFox/monaco-languageclient/actions/runs/13721043195/job/38376388610):
```shell
⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯
Canceled: Canceled
 ❯ _a45.dispose node_modules/@codingame/monaco-vscode-api/vscode/src/vs/editor/contrib/wordHighlighter/browser/wordHighlighter.js:613:22
 ❯ UniqueContainer.value node_modules/@codingame/monaco-vscode-api/vscode/src/vs/editor/contrib/wordHighlighter/browser/wordHighlighter.js:650:38

This error originated in "packages/wrapper/test/wrapper.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "Test html parameter with start". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- If the error occurred after the test had been completed, this was the last documented test before it was thrown.
```

cdietrich:
@kaisalmen tests look good. thanks for the fix

kaisalmen:
@CGNonofr I will update `@codingame/monaco-vscode-api`, test it and then ask for review/approval.

kaisalmen:
@CGNonofr this is now ready.

packages/wrapper/src/editorApp.ts
```diff
@@ -249,18 +253,30 @@ export class EditorApp {
...
             } else {
                 throw new Error('You cannot update models, because original model ref is not contained, but required for DiffEditor.');
             }
         }
+
+        monaco.editor.getModels().forEach(model => {
+            if (diposeOldModels.includes(model.uri.toString())) {
+                model.dispose();
+            }
+        });
```
CGNonofr, line 279: nit: instead of looping in each model, you can loop on your diposeOldModels and get the model by passing the url to `getModels`

btw, shouldn't you be using model ref instead and just disposing the ref?

kaisalmen: I have updated the code ⬇️

kaisalmen: Btw, is it save to compare the uris or should I compare the string representations to check if they are different. WDYT?

kaisalmen: > What do you mean? if there is no ref left, the model is disposed

Interestingly they are kept globally after disposing the reference, therefore I have to call https://github.com/TypeFox/monaco-languageclient/blob/244051d9ef8344d0df9fd4a72c46830f95cf7132/packages/wrapper/src/editorApp.ts#L275-L277 otherwise the `didClose` notification is not sent from client to server.

> Comparing uri objects using == is not safe. Why do you need to do that?

Fixed the comparison. I only want to update the editor if the uris really changed.

kaisalmen: @CGNonofr does that ⬆️ sound like something is not working correctly inside monaco?

kaisalmen: > It doesn't look safe to force dispose it I've seen some part of the code creating a reference and disposing it in a timeout

That is bad news, because I don't know how it handle it then. Btw, this works properly with the views service. This problem only occurs with the wrapper/editor service.

CGNonofr: It's no that hard to know btw, you can add some debug logs in the createModelReference function to see which ref is not released

cdietrich: to reproduce:
https://github.com/cdietrich/monaco-editor-react-new/commit/d20817c2ffc9dc450f3c6926d1dd4e2d74e540ac
and comment out
grafik

cdietrich: hmm i see createModelReference called once only

kaisalmen:
@CGNonofr @cdietrich and @pbrostean I decided to create new pre-releases, so you can test it and I can think about how to circumvent the hard model disposal:
monaco-languageclient@9.5.0-next.0
monaco-editor-wrapper@6.6.0-next.1
@typefox/monaco-editor-react@6.6.0-next.1

packages/wrapper/src/editorApp.ts
kaisalmen: @CGNonofr and @cdietrich the problem is gone. It seems that it was a combination of disposing the model reference at the wrong time and always setting the textEditorModel even if not needed. The test workarounds are also no longer required.

kaisalmen: `monaco.editor.getModel(uri)?.dispose` calls are removed.

kaisalmen: > The test workarounds are also no longer required.

They were a strong indication that I did something wrong, but I did not understand why last Friday. Anyway, enforcing everyone to deal explicitly with uris is a necessary improvement.

cdietrich:
looks good @kaisalmen

kaisalmen:
@cdietrich thank you

PR #904, opened 2025-04-03
Update complete uri and model handling in EditorApp

### LINKED ISSUE #898: Unexpected Behavior with State Changes in @typefox/monaco-editor-react 6.6.0
With the latest version of `@typefox/monaco-editor-react` (6.6.0) and the corresponding packages according to the [compatibility table](https://github.com/TypeFox/monaco-languageclient/blob/main/docs/versions-and-history.md), there is an unexpected behavior in the component when state changes occur. Previously, the following usage was not an issue:

```typescript
import { MonacoEditorReactComp } from "@typefox/monaco-editor-react"
import "index.css"
import { MonacoEditorLanguageClientWrapper, TextContents, WrapperConfig } from "monaco-editor-wrapper"
import { CSSProperties, useEffect, useMemo, useRef, useState } from "react"
import { createMonacoWrapperConfig } from "./config"

interface TestEditorProps {
    style?: CSSProperties
}

export const TestEditor = ({ style = { height: "50vH" }}: TestEditorProps) => {
    const [wrapperConfig, setWrapperConfig] = useState<WrapperConfig | null>(null)
    const [testState, setTestState] = useState<string>("")

    const monacoEditorRef = useRef<MonacoEditorLanguageClientWrapper | null>(null)
    //const memoizedWrapperConfig= useMemo(() => createMonacoWrapperConfig("Hello World!", "test.txt"), [])

    useEffect(() => {
        setWrapperConfig(createMonacoWrapperConfig("Hello World!", "test.txt"))
    }, [])

    const handleEditorLoad = (wrapper: MonacoEditorLanguageClientWrapper) => {
        monacoEditorRef.current = wrapper
    }

    const handleTextChanged = (textChanges: TextContents) => {
        if (!monacoEditorRef.current?.isStarted()) return
        setTestState(textChanges.modified as string)
    }

    return (
        <>
            {wrapperConfig && (
                <MonacoEditorReactComp
                    wrapperConfig={wrapperConfig}
                    style={style}
                    onLoad={handleEditorLoad}
                    onTextChanged={handleTextChanged}
                />
            )}
            {testState}
        </>
    )
}
```

However, since the update, all state changes lead to the following error:

```
Uncaught (in promise) Canceled: Canceled
    at Delayer.cancel (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:8208:69)
    at Delayer.dispose (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:8218:14)
    at dispose (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:2531:13)
    at _DisposableStore.clear (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:2724:11)
    at _DisposableStore.dispose (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:2714:14)
    at _a30.dispose (http://localhost:5173/node_modules/.vite/deps/chunk-2NK2GE54.js?v=d546c923:20423:23)
    at _a31.dispose (http://localhost:5173/node_modules/.vite/deps/chunk-2NK2GE54.js?v=d546c923:20494:33)
    at dispose (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:2531:13)
    at DisposableMap.clearAndDisposeAll (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:2906:11)
    at DisposableMap.dispose (http://localhost:5173/node_modules/.vite/deps/chunk-WFGYAUMR.js?v=d546c923:2899:14)
```

This is followed by a re-initialization of the editor.

Additionally, the newly introduced `automaticallyDispose` option in `wrapperConfig` does not provide the expected solution for `React.StrictMode`. The editor is still initialized and rendered twice. However, setting `automaticallyDispose` to `false` now results in the application functioning as expected, with only one instance of the editor existing (despite an error being logged in the console):

Image

kaisalmen:
Changes:

- `EditorApp` now always created models and disposes them properly when they are no longer needed. The logical structure of the code is better.
- Fix onTextChanged handling in react component. This fixes #898  => this required for changes / fixes in the `Wrapper` / `EditorApp`
- Enforce proper uri usage in tests. There were many URIs that weren't unique
- General api clean-up
- Allow model dispose in EditorApp to be asynchronous in tests. This problem does not occur in the browser at all. vitest lets tests fail due to a " Unhandled Rejection / Canceled: Canceled" error that you cannot intercept when editor models are disposed. The only way to make the tests stable is to delay the model disposal if a timeout value is larger than 0.

The react component may need further polishing depending also on other observations from people I did not yet analyse, but we should go from here...

- Fixed tests and updated vite(st) configuration with regard to dependency optimization.


packages/wrapper/src/editorApp.ts
```diff
@@ -37,6 +38,16 @@ export interface CodeResources {
     original?: CodeContent;
 }
 
+export interface CallbackDisposeable {
```
CGNonofr, line 41: Disposable :)

CGNonofr (commented):
Pretty huge commit :thinking:

packages/wrapper/src/editorApp.ts
kaisalmen: Will fix that

kaisalmen:
> Pretty huge commit 🤔

Yes, but model/uri handling was still fishy and that made this change bigger than initially expected. vitest gave me some headache as well.

packages/wrapper/src/editorApp.ts
kaisalmen: Force pushed. I found some other left-overs ⬇️

kaisalmen:
@CGNonofr any thoughts? I have released next versions based on the current status, btw.

CGNonofr:
> @CGNonofr any thoughts? I have released next versions based on the current status, btw.

I'm sorry, I don't have as much time I'd like to be able to fully understand the changes you made. I'm sure it's for the best!

kaisalmen:
@CGNonofr take your time. People can provide feedback based on the next packages. I also want to do some react component testing myself.

packages/wrapper-react/src/index.tsx
```diff
@@ -49,7 +49,7 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
         const startMonaco = async () => {
             if (containerRef.current) {
                 try {
-                    wrapperRef.current.registerTextChangeCallback(onTextChanged);
+                    wrapperRef.current.registerTextChangedCallback(onTextChanged);
```
zoran995, line 52: The new dependency array will always preserve the old reference of onTextChanged, can we create a callback and wrap the onTextChanged
i.e.
```tsx
const onTextChangedRef = useRef(onTextChanged);
onTextChangedRef.current = onTextChanged;
```
and here
```tsx
wrapperRef.current.registerTextChangedCallback(() => onTextChangedRef.current());
```
this way we register only once but always call the new callback (users also don't need to put onTextChanged in useCallback)

kaisalmen: Yes, thank you, that sounds like a good idea.

kaisalmen: Done,. see ⬇️

zoran995:
Hi @kaisalmen I just tried upgrading to 6.7.0-next.0 and it still throws error in strict mode
```
Uncaught (in promise) Error: file 'extension-file://typefox.statemachine-example/extension/react-statemachine-configuration.json/' already exists
```  
and I also get ```Extension '.liquid-example` is already registered```, will try to figure out the reproduction steps, but posting as you might have some idea

zoran995:
@kaisalmen I still believe there is no straightforward way to do this without implementing a task queue. Everything is asynchronous, and although we expect the component to initialize first and then dispose, we can't guarantee this sequence with asynchronous initialization and disposal.
Since useEffect doesn't support asynchronous functions, we need to make everything appear synchronous, which creates a problem in strict mode as React receives the useEffect callback before the initialization is completed. After receiving the callback, React expects that all work is completed and calls it, but in our case, it causes issues as the component doesn't know it should be destroyed and continues with initialization and starting.
With the task queue, we can ensure that the sequence of initialization and disposal is completed one after another, rather than in parallel. Basically you should be able to use the entire implementation from https://github.com/TypeFox/monaco-languageclient/pull/872. I just tested it and couldn't reproduce the initialization issue; it seems to be fixed with something else.

kaisalmen:
Hi @zoran995 I cannot reproduce the behaviour with a unit test, yet. I have locally defined another one that uses strict mode. I need to see if I make web worker language server run in a test (it did not work in the past if I remember correctly).

I am hesitant to introduce something complex to overcome a "dev" feature. I am not implying here that what it shows is wrong, but I am favoring a simpler solution if possible. There must have been something else been broken during the "simplification" in February (before `6.4.0` release) and I have an idea: You weren't able to restart during init and start combined. Now it is possible to do that I guess that introduced issues. I will come back...

kaisalmen:
@CGNonofr and @zoran995 new preview builds (`-next.1`) are available.

CGNonofr (approved):

PR #915, opened 2025-04-24
Fix data URL crash for Unicode characters

metp:
Fixes `DOMException: String contains an invalid character` in `verifyUrlOrCreateDataUrl` when a data url is created of a string that contains a Unicode character larger than one byte.

See https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa#unicode_strings for a detailed explanation.

I had a special Unicode character in my language configuration which caused this error. Looks like the receiving side can deal with the UTF8 encoded string as it works as expected for me now.

### DIFF
--- a/packages/wrapper/test/editorApp.test.ts
+++ b/packages/wrapper/test/editorApp.test.ts
@@ -22,7 +22,10 @@ describe('Test EditorApp', () => {
     test('verifyUrlorCreateDataUrl: url', async () => {
         const url = new URL('../../../node_modules/langium-statemachine-dsl/syntaxes/statemachine.tmLanguage.json', window.location.href);
         const text = await (await fetch(url)).text();
-        expect(verifyUrlOrCreateDataUrl(text)).toBe(`data:text/plain;base64,${btoa(text)}`);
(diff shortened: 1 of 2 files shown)


packages/wrapper/test/editorApp.test.ts
```diff
@@ -22,7 +22,10 @@ describe('Test EditorApp', () => {
     test('verifyUrlorCreateDataUrl: url', async () => {
         const url = new URL('../../../node_modules/langium-statemachine-dsl/syntaxes/statemachine.tmLanguage.json', window.location.href);
         const text = await (await fetch(url)).text();
-        expect(verifyUrlOrCreateDataUrl(text)).toBe(`data:text/plain;base64,${btoa(text)}`);
+        const bytes = new TextEncoder().encode(text);
```
kaisalmen, line 25: Can you add a new unit test that verifies that Unicode characters now work?

kaisalmen (commented):
Hi @metp thank you. Can you please add another unit test that verifies this issue with Unicode characters is properly solved?

kaisalmen (approved):
Thank you very much, LGTM

PR #918, opened 2025-05-05
Always create a getWorker function

### LINKED ISSUE #913: [Bug] Monaco-Editor-React Fails Without Default monacoWorkerFactory
Setting up the latest [@typefox/monaco-editor-react](https://www.npmjs.com/package/@typefox/monaco-editor-react?activeTab=readme) (6.6.0, and 6.7.0-next.1) leads to an issue where some workers are unavailable to the editor (editorWebWorker & textMateTokenizationWorker). Here's a quick image of the observed errors below.

Image

Since the modules can't be loaded up, the whole setup appears to stop at that point. Leaving the editor intact, but without any applied highlighting or LSP functionality from an associated LS worker.

The problem appears to stem from not having the default worker factory in place for the `monacoWorkerFactory` config option.
```ts
const createDefaultWrapperConfig = (): WrapperConfig => {
    return {
      $type: "extended",
      logLevel: LogLevel.Debug,
      editorAppConfig: {
        codeResources: {
          modified: {
            text: 'print("Hello, World!")',
            uri: "/workspace/hello.py",
          },
        },
        // applying the default worker factory here resolves the issue
        //monacoWorkerFactory: configureDefaultWorkerFactory,
      },
    };
}
```

The addition of the monacoFactoryWorker is notably not present on the README, which means the current default setup won't work as expected without modification. It appears that previously this worked just fine, but an upstream change looks to have broken this behavior in some way.

Associated `@codingame/monaco-...` packages are all at 16.0.4 for reference.

[Here's a minimal project setup ](https://github.com/montymxb/monaco-editor-react-test/tree/broken-example) using a fresh Vite + TS + React to demonstrate this using a slightly adapted version of the default Python example on the README. The issue is setup on the 'broken-example' branch as a heads-up, and the snippet above can be found in **App.tsx**.

kaisalmen:
This fixes #913 

A `getWorker` function is always created. If none is provided with the `WrapperConfig` an empty one is created. The empty one will throws errors for all requested workers.

### DIFF
--- a/packages/client/src/vscode/services.ts
+++ b/packages/client/src/vscode/services.ts
@@ -18,6 +18,7 @@ import type { EnvironmentOverride } from '@codingame/monaco-vscode-api/workbench
 import type { Logger } from 'monaco-languageclient/tools';
 import { FakeWorker as Worker } from './fakeWorker.js';
 import { setUnexpectedErrorHandler } from '@codingame/monaco-vscode-api/monaco';
+import { useWorkerFactory } from '../workerFactory.js';
 
 export interface MonacoEnvironmentEnhanced extends monaco.Environment {
     vscodeInitialising?: boolean;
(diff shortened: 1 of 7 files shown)


CGNonofr (approved):
LGTM

Issue #919, opened 2025-05-05
Separate language client (configuration) from wrapper and react component


kaisalmen:
These are some ideas I want to share:

The MonacoLanguageClient Wrapper is basically an independent thing now and it should be moved from the wrapper to the monaco-languageclient package. The specfic configuration objects should be moved there as well. The wrapper configuration should be split-up and the life-cycles should of wrapper and languageclients should be separated especially with regard to the react component.

It became clear in the last couple of weeks that the global configuration items (service init, worker config and the language clients could be seen as such) should be separated from the editor related life-cycle. This is not such a much a problem for the wrapper, but for the react component. The whole stack cannot be treated like a regular component (it is a hydra in a box). Some things can be re-initialized, but others just can't and should not. This must be clear by the architecture and the API of languageclient, wrapper and the react component.

@CGNonofr I will update this further tomorrow.

kaisalmen:
@CGNonofr I have re-worked the text. I hope this is clearer now. WDYT?

kaisalmen:
@CGNonofr I am getting closer to release a new next version. This will be a major version and all `monaco-editor-wrapper` features moved to the monaco-languageclient package as sub-package exports (`monaco-editor-wrapper` will be obsolete). Everything regarding `monaco-vscode-api` handling is gathered in one place. Only the react component is kept separate, because I don't want to have react as external dependencies to `monaco-languaggeclient`.  `vscode-ws-jsonrpc` is still there.

I need to do a full write-up at some point what changed and why. And I would also like you at least (everyone else is welcome of course) to test the new version and look at the changes. I use a test driven approach to analyse/overcome the react issues and that so far has proven a very good approach. vitest browser mode + playwright is superb.

Ranjit-Khanal:
@kaisalmen  I'm using   "@typefox/monaco-editor-react": "^6.9.0",  "monaco-editor-wrapper": "^6.9.0",
    "monaco-languageclient": "^9.8.0",

Ranjit-Khanal:
@kaisalmen `` I'm working on integrating the monaco-languageclient with @monaco-editor/react (the React wrapper for Monaco) in my project. I’ve been exploring how to make the monaco-languageclient automatically handle LSP features such as hover, go to definition, diagnostics, and completion, but I’m running into integration challenges — especially with the wiring between the Monaco editor instance and the language client setup.

So far, I've done the following:

Rendered the Monaco editor using @monaco-editor/react.

[...]

PR #920, opened 2025-05-06
Move MonacoLanguageClientWrapper to monaco-languageclient

kaisalmen:
This just movse `MonacoLanguageClientWrapper` including config and tests to `monaco-languageclient`. It does not change any beahviour, yet.

It introduces a sub-export `monaco-languageclient/warpper` and increase the version to `10.0.0-next.0`.

This branch is merged to the dev branch and it will contain all the changes needed for #919 until being merged back to main.


CGNonofr:
> @CGNonofr this makes` monaco-languageclient` dependent on `vscode-ws-jsonrpc` and we did not wanted that in the past. So, maybe this is not a good idea. WDYT?

Why didn't we want that ? :thinking:

kaisalmen:
> Why didn't we want that ? 🤔

Good question. Maybe, I did not remember that correctly

kaisalmen:
@CGNonofr ok, I made the adjustment. Build should be green now.

CGNonofr (approved):

PR #949, opened 2025-08-04
Use public exports from vscode-jsonrpc instead of deep internal imports

masrur-mmrs:
### Problem
`vscode-ws-jsonrpc` was importing internal modules of `vscode-jsonrpc` such as:
  import { AbstractMessageReader } from 'vscode-jsonrpc/lib/common/messageReader.js';

With the introduction of an "exports" field in `vscode-jsonrpc` (Which was another problem in and of itself, here's the [issue](https://github.com/microsoft/vscode-languageserver-node/issues/1662)), consumers
can no longer reach into those internal paths, causing errors like:
 `Error: Package subpath './lib/common/messageReader.js' is not defined by "exports"`
Also, the existing TypeScript config did not consistently opt into
modern module resolution modes needed to work with exported subpaths.

### Solution
- Replaced all deep/internal `vscode-jsonrpc/lib/...` imports with public exports
  (e.g., `import { AbstractMessageReader } from 'vscode-jsonrpc'`).
- Adjusted `tsconfig.src.json` to enable `module: NodeNext`, `moduleResolution: NodeNext`,
  and enabled `declaration` output so the build can resolve the exported surface and
  emit type definitions correctly.
- Ensures compatibility with strict package export policies and avoids relying on
  implementation details.

### DIFF
--- a/packages/vscode-ws-jsonrpc/src/socket/reader.ts
+++ b/packages/vscode-ws-jsonrpc/src/socket/reader.ts
@@ -4,6 +4,8 @@
  * ------------------------------------------------------------------------------------------ */
 
 import { Disposable } from 'vscode-jsonrpc';
+// TODO: Use environment-specific imports (vscode-jsonrpc/browser or vscode-jsonrpc/node)
+// when upgrading to vscode-jsonrpc@9.x.x-next.X which supports proper export maps
 import { type DataCallback, AbstractMessageReader, MessageReader } from 'vscode-jsonrpc';
 import type { IWebSocket } from './socket.js';
 
--- a/packages/vscode-ws-jsonrpc/src/socket/reader.ts
+++ b/packages/vscode-ws-jsonrpc/src/socket/reader.ts
@@ -4,7 +4,7 @@
  * ------------------------------------------------------------------------------------------ */
 
 import { Disposable } from 'vscode-jsonrpc';
-import { type DataCallback, AbstractMessageReader, MessageReader } from 'vscode-jsonrpc/lib/common/messageReader.js';
+import { type DataCallback, AbstractMessageReader, MessageReader } from 'vscode-jsonrpc';
 import type { IWebSocket } from './socket.js';
 
 export class WebSocketMessageReader extends AbstractMessageReader implements MessageReader {
--- a/packages/vscode-ws-jsonrpc/src/socket/writer.ts
+++ b/packages/vscode-ws-jsonrpc/src/socket/writer.ts
@@ -3,8 +3,8 @@
  * Licensed under the MIT License. See LICENSE in the package root for license information.
  * ------------------------------------------------------------------------------------------ */
 
-import { Message } from 'vscode-jsonrpc/lib/common/messages.js';
-import { AbstractMessageWriter, MessageWriter } from 'vscode-jsonrpc/lib/common/messageWriter.js';
+import { Message } from 'vscode-jsonrpc';
+import { AbstractMessageWriter, MessageWriter } from 'vscode-jsonrpc';
 import type { IWebSocket } from './socket.js';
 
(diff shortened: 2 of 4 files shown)


CGNonofr:
AS per the version 8.2.1

image

There is no default `browser` field, so won't importing `vscode-jsonrpc` directly be importing node stuff in the browser?

masrur-mmrs:
> AS per the version 8.2.1
> image
> 
> There is no default `browser` field, so won't importing `vscode-jsonrpc` directly be importing node stuff in the browser?

True, there is no default browser field, which is why I added a "browser" field in my vscode-jsonrpc 
```
"exports": {
		".": {
			"types": "./lib/common/api.d.ts",
			"default": "./lib/common/api.js"
		},
		"./node": {
			"types": "./lib/node/main.d.ts",
			"node": "./lib/node/main.js"
		},
		"./browser": {
			"types": "./lib/browser/main.d.ts",
			"browser": "./lib/browser/main.js"
		}
	},
```
I would like to clarify, though, that my issue was closed. The problem was that only the new insider version of `vscode-jsonrpc @9.0.0-next.8` supports entry points, but `vscode-ws-jsonrpc` uses `vscode-jsonrpc @8.2.1`

To get things working on my end I had to compile my own versions of both `vscode-jsonrpc` and `vscode-ws-jsonrpc`

masrur-mmrs:
Actually I probably should have done this instead:
```json
"exports": {
		".": {
			"import": "./lib/esm/main.js",
			"require": "./lib/umd/main.js",
			"node": "./lib/node/main.js",
			"default": "./lib/node/main.js",
			"browser": "./lib/browser/main.js"
		},
		"./node": {
			"import": "./lib/esm/node.js",
			"require": "./lib/umd/node.js"
		},
		"./browser": {
			"import": "./lib/esm/browser.js",
			"require": "./lib/umd/browser.js"
		}
	},
```
I actually didn't think about the fact that without proper browser entry points I'm risking pulling node code into browser bundles. All I wanted to do was to get both of these to work together.

masrur-mmrs:
> Thank you, can you drop the tsconfig changes? This is all driven by the base config. I propose to release a next release after this PR is merged, so you can test the improvements.

Absolutely, I'll also add a comment to remind environment specific imports when the next release of `vscode-jsonrpc` is used

masrur-mmrs:
The failed worker tests seem to be unrelated to the changes I made to the imports under `vscode-ws-jsonrpc/socket`
Interestingly enough, all the tests pass in my local environment.

kaisalmen:
> Interestingly enough, all the tests pass in my local environment.

I will re-run the test.

kaisalmen:
There are some time out issues with the tests that happen once in a while. It will improve in the future (dev branch).

kaisalmen:
@masrur-mmrs I included all your changes in: https://github.com/TypeFox/monaco-languageclient/tree/update-vwj

masrur-mmrs:
> @masrur-mmrs I included all your changes in: https://github.com/TypeFox/monaco-languageclient/tree/update-vwj

Appreciate it

kaisalmen:
@masrur-mmrs next versions are available now:
https://www.npmjs.com/package/vscode-ws-jsonrpc/v/3.5.0-next.0
https://www.npmjs.com/package/monaco-languageclient/v/9.9.0-next.0

masrur-mmrs:
> @masrur-mmrs next versions are available now: https://www.npmjs.com/package/vscode-ws-jsonrpc/v/3.5.0-next.0 https://www.npmjs.com/package/monaco-languageclient/v/9.9.0-next.0

Everything works without errors now. Also thanks for including the recommendation to use `monaco-editor-vscode-api` as an alias to `monaco-editor`  
Screenshot 2025-08-09 at 5 37 14 PM
This significantly helps in building an OCE with vscode features most people expect.

PR #954, opened 2025-09-10
Improvements after initial doc editing

kaisalmen:
Most changes correct api inconsistencies:

- apiWrapper: Make ViewsConfig config mandatory.
  - Hereby clarify react usage (HtmlContainerConfig) and later override before start in react
  - Move htmlElement into ViewsConfig
- Rename MonacoVscodeApiConfigRuntime method init to start
- Remove $type from EditorAppConfig, so it is only configured in a single place

There are two bug fixes for the react component:

- Now all async updates after global init are put into a queue, so the execution order is ensured.
- Only one language client can be configured directly. More complex things must be done at a potential application level because language client life cycle should not be bound to the react component. That you configure it is just a convenience for simple applications.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -26,6 +26,7 @@ pluginHeader.rules.header.meta.schema = false;
 
 export default [{
     ignores: [
+        '**/.chrome/**/*',
         '**/node_modules/**/*',
         '**/dist/**/*',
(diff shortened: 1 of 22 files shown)


CGNonofr (approved):
I've seen nothing weird but I can say I've understood every change!

PR #964, opened 2025-09-29
tsgo and erasableSyntaxOnly

kaisalmen:
tsgo is stable for our use-case
`erasableSyntaxOnly` is now set in tsconfig. The changes required are not severe.

I guess we will have a `10.1.0` release soon.


kaisalmen:
I already added the dependency a couple of month ago, but it did not support `--build` in the beginning.

PR #970, opened 2025-10-06
fix example in client README

IoannisNezis:
Hey there,

I updated to monaco-languageclient and noticed this error in the example.

Otherwise the migration went smooth, thanks a lot!


CGNonofr:
Is it an error? this syntax is supported

PR #977, opened 2025-10-20
Move generic debugger client code to mlc

kaisalmen:
It does not hurt to move the general debugging client connection code from examples to the client


CGNonofr (approved):
LGTM

PR #986, opened 2025-11-03
Fixed re-rendering issues with React StrictMode

### LINKED ISSUE #975: Rerenders cause editor element to be recreated for MonacoEditorReactComp
I created a reproducible example with a purple button where every time you click it will update the text and cause another div to be created:

https://github.com/KamWithK/monaco-languageclient-demo

### LINKED ISSUE #988: Initialization of vscode services can only performed once!
When clicking mount, the client was rendered normally for the first time, and when clicking uninstall, it was also uninstalled normally. However, when clicking mount again, the client could not be rendered normally and the console reported an error
``Initialization of vscode services can only performed once! ``
Can it only be mounted once? Have you considered the scenario of needing to uninstall and then remount?
```TypeScript
/* --------------------------------------------------------------------------------------------
 * Copyright (c) 2024 TypeFox and others.
 * Licensed under the MIT License. See LICENSE in the package root for license information.

import * as vscode from 'vscode';
import { type RegisterLocalProcessExtensionResult } from '@codingame/monaco-vscode-api/extensions';
import React, { useState } from 'react';
import ReactDOM from 'react-dom/client';
import { MonacoEditorReactComp } from '@typefox/monaco-editor-react';
import { MonacoEditorLanguageClientWrapper } from 'monaco-editor-wrapper';
import { createWrapperConfig } from './config.js';
import { configureDebugging } from '../../debugger/client/debugger.js';

export const runPythonReact = async () => {
    const appConfig = createWrapperConfig();
    
    const onLoad = async (wrapper: MonacoEditorLanguageClientWrapper) => {
        const result = wrapper.getExtensionRegisterResult('mlc-python-example') as RegisterLocalProcessExtensionResult;
        result.setAsDefaultApi();

        const initResult = wrapper.getExtensionRegisterResult('debugger-py-client') as RegisterLocalProcessExtensionResult | undefined;
        if (initResult !== undefined) {
            configureDebugging(await initResult.getApi(), appConfig.configParams);
        }

        await vscode.commands.executeCommand('workbench.view.explorer');
        await vscode.window.showTextDocument(appConfig.configParams.files.get('hello2.py')!.uri);
    };

    const root = ReactDOM.createRoot(document.getElementById('react-root')!);

    const App = () => {
        const [dispose, setDispose] = useState(false)
        return (
            <div style={{ 'backgroundColor': '#1f1f1f' }} >
                <div style={{ display: 'flex' }}>
                    <button onClick={() => setDispose(true)}>mount</button>
                    <button onClick={() => setDispose(false)}>unmount</button>
                </div>
                {
                    dispose ?
                        <MonacoEditorReactComp
                            wrapperConfig={appConfig.wrapperConfig}
                            style={{ 'height': '100%' }}
                            onLoad={onLoad}
                            onError={(e) => {
                                console.error(e);
                            }} /> : null
                }
            </div>
        );
    };
    root.render(<App />);
};
```

kaisalmen:
- Renders editor only once
- Updated tests
- Enforced language client disposal was moved from config to the react component (updated existing test and added StrictMode specific test)
- Moved StrictMode specific test to own file
- Renamed LanguageClientsManager to LanguageClientManager
- Update dependencies
- Fix lint issues

Fixes #975 
Fixes #988

### DIFF
--- a/packages/wrapper-react/src/index.tsx
+++ b/packages/wrapper-react/src/index.tsx
@@ -35,7 +35,7 @@ const lcsManager = new LanguageClientManager();
 const haveEditorService = () => {
     return getEnhancedMonacoEnvironment().viewServiceType === 'EditorService';
 };
-const logger = new ConsoleLogger(LogLevel.Debug);
+const logger = new ConsoleLogger(LogLevel.Off);
 
 const runQueue: Array<{id: string, func: () => Promise<void>}> = [];
 let runQueueLock = true;
@@ -50,8 +50,6 @@ const addQueue = (id: string, func: () => Promise<void>) => {
 };
 
 const executeQueue = async () => {
-    console.log(`Queue size: ${runQueue.length}`);
-
     if (runQueue.length > 0) {
         runQueueLock = true;
         while (runQueue.length > 0) {


kaisalmen:
@CGNonofr I will release a next versions soon, so people can test it. If that looks good, the PR will be ready for review.

packages/wrapper-react/src/index.tsx
```diff
@@ -47,31 +52,47 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
+    const triggerQueue = () => {
+        setInterval(() =>  {
+            if (queueAwait === undefined) {
+                queueAwait = new Promise<void>((resolve) => {
+                    queueResolve = resolve;
+                });
+                executeQueue();
+            }
+        }, 50);
+    };
```
CGNonofr: I'm confused by that code :thinking: 

- it uses a local variable which is destroyed at each render
- nothing stops the interval :thinking:

packages/wrapper-react/src/index.tsx
```diff
@@ -132,25 +147,41 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
+        // fast-fail
+        if (editorAppConfig === undefined) return;
+
         // always try to perform global init. Reason: we cannot ensure order
         performGlobalInit();
 
+        let createEditor = false;
+        // it is possible to run without an editorApp, for example when using the ViewsService
+        if (haveEditorService.current) {
+            createEditor = currentEditorConfig.current === undefined || JSON.stringify(editorAppConfig) !== JSON.stringify(currentEditorConfig.current);
```
CGNonofr: I'm not very comfortable with comparing the old and new config by value like that :thinking: because if the configuration is generated in the caller component render, it will change every times, and we'll serialize it at each render as well to compare it without any warning :thinking: 

Also using JSON.stringify to do that is probably not the more robust way :thinking: 

At least, we can compare the ref before comparing the value I guess

kaisalmen:
@CGNonofr this is ready to review now. I added additional react component unit tests We now have 20 unit tests for the react component and 111 total. The new test cases ensure the fixes for #975 and #988 are valid and that the fixes are not invalidated in the future. vite browser tests rule! 🙂 

@martin-fleck-at do see any further problems on your end?

packages/wrapper-react/src/index.tsx
kaisalmen: Oh damn, you are right. If I move this out of the component the interval is only kicked once. I just tested it locally, but it runs forever.

kaisalmen: I have changed the implementation. It is now kicked by the addQueue function and invokes itself until the queue is depleted.

packages/wrapper-react/src/index.tsx
```diff
@@ -36,8 +37,46 @@ const haveEditorService = () => {
...
+        // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+        queueAwait?.resolve();
+        queueAwait = undefined;
+        executeQueue();
+    }
+};
+
+const debugLogging = (id: string, useTime?: boolean) => {
+    if (useTime === true) {
+        apiWrapper?.getLogger().debug(`${id}: ${Date.now()}`);
```
martin-fleck-at: nit: I noticed that I can set the debug level on the editor app configuration but it is not considered by this logger. Furthermore, if I do not provide the `apiWrapper` because I already initialized the API, nothing will be logged at all. Maybe it would be nice for the editor to have it's own `ConsoleLogger` like the editor app configuration.

(just a minor thing I noticed but not really part of this PR so please feel free to ignore)

kaisalmen: Yes, that could make sense. Also, the language client could have its own logger. This makes generally sense if you want to have different log-levels for the different things.

packages/wrapper-react/src/index.tsx
kaisalmen: I don't have a good answer here right now. I am playing around. The dual-editors in StrictMode are back. I am really thinking that only changes to code and editor configuration should be dynamic. If you are not using react, you will only destroy the editorApp if you fundamentally change something, I think.

packages/wrapper-react/src/index.tsx
kaisalmen: @CGNonofr ok, now the queue is kicked and stopped properly and the double-rendering is gone, because the previous fix was not good as the queue was entered multiple times.

kaisalmen:
@CGNonofr and @martin-fleck-at I converted it back to a Draft. I want to fix the problems you mentioned, first. This is getting lengthy, in kn ow, but theses are all valid points. The first issue @CGNonofr brought up is resolved, though.

packages/wrapper-react/src/index.tsx
kaisalmen: It is finally working in all cases. `codeResource` changes and monaco options updates are passed to monaco-editor and they don't require a restart of the editor.

packages/wrapper-react/src/index.tsx
kaisalmen: Done ⬇️

kaisalmen:
@CGNonofr rework is complete. Ready for review, Part 2. New next releases will be there soon.

kaisalmen:
These new next releases are in-line with the last commit:
https://www.npmjs.com/package/monaco-languageclient/v/10.3.0-next.3
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.3.0-next.3

kaisalmen:
115 tests! 🎉

kaisalmen:
@CGNonofr this seems really stable now. The queue crash because the called functions threw exceptions in StrictMode I didn't see. There now is a catch the changes the flag and throws the exceptions. I removed some extra properties. Now the component properly handles dispose of the languageclient if enforced via the config. Code updates in `editorAppConfig` are properly processed.
New preview releases are out:
https://www.npmjs.com/package/monaco-languageclient/v/10.3.0-next.4
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.3.0-next.4

packages/client/src/editorApp/config.ts
```diff
@@ -52,8 +52,8 @@ export interface EditorAppConfig {
     domReadOnly?: boolean;
     readOnly?: boolean;
     overrideAutomaticLayout?: boolean;
-    editorOptions?: monaco.editor.IStandaloneEditorConstructionOptions;
-    diffEditorOptions?: monaco.editor.IStandaloneDiffEditorConstructionOptions;
+    editorOptions?: monaco.editor.IStandaloneEditorConstructionOptions | monaco.editor.IEditorOptions;
```
CGNonofr: What is the point of this change? IStandaloneEditorConstructionOptions already extends IEditorConstructionOptions which extends IEditorOptions :thinking: (and all fields are optional)

packages/wrapper-react/src/index.tsx
```diff
@@ -38,120 +111,79 @@ export const MonacoEditorReactComp: React.FC<MonacoEditorProps> = (props) => {
...
+                        });
+                    }
+                    await apiWrapper.start();
+                    onVscodeApiInitDone?.(apiWrapper);
 
-                onVscodeApiInitDone?.(apiWrapperRef.current);
+                    debugLogging('GLOBAL INIT DONE', true);
 
-                debugLogging('GLOBAL INIT DONE', true);
+                    queueLock = false;
```
martin-fleck-at: I just tried this but in my case where the Monaco API is already initialized, the queue is never actually being executed since the `queueLock` is never set to false. So what I end up seeing in the log is:

```
DEBUG Adding to queue: editorInit: QUEUE SIZE before: 0
DEBUG Adding to queue: dispose: QUEUE SIZE before: 1
DEBUG Adding to queue: editorInit: QUEUE SIZE before: 2
DEBUG Checking queue...true
DEBUG Checking queue...true
DEBUG Checking queue...true
DEBUG Checking queue...true
DEBUG Checking queue...true
DEBUG Checking queue...true
DEBUG Checking queue...true
...
```

I think the queue needs to be unlocked as soon as Monaco is done initializing or if it is already initialized.

packages/client/src/editorApp/config.ts
kaisalmen: Right, it is actually not needed. I don't recall and reverted it already locally.

packages/wrapper-react/src/index.tsx
kaisalmen: @martin-fleck-at thanks. I fixed that and duplicated all basic react tests to use an independently started monaco-vscode-api.
Error handling runs through the same function for all code paths now. queue is stopped if there is no handler configured.
New version will be there soon.

kaisalmen:
@CGNonofr and @martin-fleck-at thanks for the feedback. Things are addressed. New preview builds here:
https://www.npmjs.com/package/monaco-languageclient/v/10.3.0-next.5
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.3.0-next.5

martin-fleck-at:
@kaisalmen Thank you very much for the update, everything works so much stabler now! I do experience some issues in my use case: I do get the initial content from external and add it as part of the modified code resource text. Then I have a listener that tracks text changes and updates the external system again which will trigger another update in the rendering of the editor with the new value. So it is something like this:

```
  const onTextChanged = useCallback(
    (change: { modified?: string }) => {
      onChange(change.modified ?? ''); // notify external system
    },
    [onChange]
  );

  const editorAppConfig = useMemo(() => {
    return {
      codeResources: {
        modified: {
          uri: ...,
          text: value,
          enforceLanguageId: language
        }
      },
      [...]
    };
  }, [language, value, ...]); // language and value are handed it from this wrapper component

<MonacoEditor
    editorAppConfig={editorAppConfig}
    onTextChanged={onTextChanged}
    ...
/>
```

So value gets updated on every key stroke but with a delay which leads to the cursor being reset all the time for me:
editor-issue

Might be something I need fix on my side but I was just wondering whether that would be the correct way to use the editor with an external system.

kaisalmen:
~~@martin-fleck-at sorry, fixed so many things, but overlooked that I now broke the editor in the component.~~

Nevermind, just a bug in the Statemachine example.

kaisalmen:
> Might be something I need fix on my side but I was just wondering whether that would be the correct way to use the editor with an external system.

@martin-fleck-at  only one actor should change the text. Maybe we can decouple this problem from the PR and analyse that independently.

Regarding my above statement I need to check if I only trashed the example or really broke something.

kaisalmen:
@martin-fleck-at I can reproduce [the issue](https://github.com/TypeFox/monaco-languageclient/pull/986#issuecomment-3527538404) It is related to my changes to -next.4 from last night. The Statemachine example suffers from the same root cause...

martin-fleck-at:
@kaisalmen I don't see any problem in splitting out particular issues, especially if I'm the only one suffering from them. Unfortunately, it seems that in this case, it is not like that.

kaisalmen:
Hi @martin-fleck-at just for confirmation: I created a problem with next.4 that resulted in the problem you observed. I removed some code that shouldn't have been removed late last night. So, this must be fixed with this PR for sure.

Additionally, the jumping gets more likely due to async handling => code / text editor update must be handled synchronously. I have something promising locally, but some units test fail now. New version likely late today or tomorrow. AFK for a bit now.

kaisalmen:
@CGNonofr please have another look. @martin-fleck-at the text editing experience should be better now. If you type very, very fast, the cursor jumps to the start otherwise this looks good. Tests are aligned. The Statemachine example works fine again as well.

New preview releases are available:
https://www.npmjs.com/package/monaco-languageclient/v/10.3.0-next.6
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.3.0-next.6

kaisalmen:
@CGNonofr and @martin-fleck-at I hope this is now good. It has been a far longer journey than I initially expected. Thank you for the constructive and helpful feedback.

If there is nothing blocking, I want to merge this and create a new final release soon (using latest monaco-vscode-api, too). There will be new issue for sure 😎 but we should try to process smaller chunks next time. 😊

martin-fleck-at:
@kaisalmen Thank you for not giving up on this! I know it turned out to be much larger than expected, I appreciate it!

kaisalmen:
Final releases are available since yesterday:
https://www.npmjs.com/package/monaco-languageclient/v/10.3.0
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.3.0

PR #997, opened 2025-11-18
Use flags to trigger behaviour

### LINKED ISSUE #996: [React Editor] Very fast typing may reset the curser and insert text incorrectly
This is extracted from https://github.com/TypeFox/monaco-languageclient/pull/986#issuecomment-3529886336 where it is noted that if you type very fast, the cursor might be reset and the text may be inserted incorrectly. The original issue has been improved a lot upon (thank you!) but it is still possible to hit this - either manually if you try or easily during tests.

editor-issue

I open this bug to keep track of the issue as testing is an important scenario for us.

kaisalmen:
Use flags `reprocessConfig` to trigger a editor config re-processing. This can trigger an on-demand re-processing of the text/model and editor options (test with statemachine react example)
Use `enforceLanguageClientDispose` to dispose the language client on demand: This was moved from the config to the react component as it is only used here.
Debug logging was improved.
This resolves #996

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11504,7 +11504,7 @@
     },
     "packages/client": {
       "name": "monaco-languageclient",
-      "version": "10.4.0-next.1",
+      "version": "10.4.0-next.2",
       "license": "MIT",
       "dependencies": {
         "@codingame/monaco-vscode-api": "^23.1.0",
@@ -11556,7 +11556,7 @@
     },
     "packages/examples": {
       "name": "monaco-languageclient-examples",
-      "version": "2025.11.2-next.1",
+      "version": "2025.11.2-next.2",
       "license": "MIT",
       "dependencies": {
         "@codingame/monaco-vscode-configuration-service-override": "^23.1.0",
@@ -11594,12 +11594,12 @@
         "@codingame/monaco-vscode-view-status-bar-service-override": "^23.1.0",
         "@codingame/monaco-vscode-view-title-bar-service-override": "^23.1.0",
         "@codingame/monaco-vscode-views-service-override": "^23.1.0",
-        "@typefox/monaco-editor-react": "~7.4.0-next.1",
+        "@typefox/monaco-editor-react": "~7.4.0-next.2",
         "cors": "~2.8.5",
         "express": "~5.1.0",
         "jszip": "~3.10.1",
         "langium": "~4.1.0",
-        "monaco-languageclient": "~10.4.0-next.1",
+        "monaco-languageclient": "~10.4.0-next.2",
         "pyright": "~1.1.407",
         "react": "~19.2.0",
         "react-dom": "~19.2.0",
@@ -11641,7 +11641,7 @@
     },
     "packages/wrapper-react": {
       "name": "@typefox/monaco-editor-react",
-      "version": "7.4.0-next.1",
+      "version": "7.4.0-next.2",
       "license": "MIT",
       "dependencies": {
         "@codingame/monaco-vscode-editor-api": "^23.1.0",
--- a/packages/client/CHANGELOG.md
+++ b/packages/client/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this npm module are documented in this file.
 
-## [10.4.0-next.1] - 2025-11-20
+## [10.4.0-next.2] - 2025-11-20
 
 - Use flags to trigger behaviour [#997](https://github.com/TypeFox/monaco-languageclient/pull/997).
 - Updated all `@codingame/monaco-vscode` packages to `23.1.0`.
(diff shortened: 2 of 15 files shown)


kaisalmen:
@CGNonofr yes, forgot to mention. The debug logging is improved (better timestamp and better and more precise messages).

kaisalmen:
A new next release is available (only the component this time):
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.4.0-next.0

kaisalmen:
@martin-fleck-at is this a possible solution for you?

martin-fleck-at:
@kaisalmen I am a bit swamped today. I might have a chance to try this out but I might only be able to have a more detailed look tomorrow. Sorry for the delay!

kaisalmen:
@martin-fleck-at now worries. Take your time. I will wait for you feedback.

kaisalmen:
New next releases are available:
https://www.npmjs.com/package/monaco-languageclient/v/10.4.0-next.1
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.4.0-next.1

packages/wrapper-react/src/index.tsx
```diff
@@ -89,12 +91,9 @@ const stopQueue = () => {
...
 
-const debugLogging = (id: string, useTime?: boolean) => {
-    if (useTime === true) {
-        logger.debug(`${id}: ${Date.now()}`);
-    } else {
-        logger.debug(id);
-    }
+const debugLogging = (id: string) => {
+    const now = new Date(Date.now());
+    logger.debug(`[${now.getMinutes()}:${now.getSeconds()}:${now.getMilliseconds()}]: ${id}`);
```
martin-fleck-at: nit: I'm a big fan of this more detailed logging approach but I feel it is a bit uncommon to see the timestamp without the hour and still formatted like a typical time, e.g.:
```
log.js:155 DEBUG [19:0:614]: Adding to queue: editorInit: QUEUE SIZE before: 0
```
Maybe some improvements:
- Add hour 
- Show hours, minutes and seconds with two digits
- Use a `.` separator (as opposed to `:`) between seconds and milliseconds as it is more common

packages/wrapper-react/src/index.tsx
```diff
@@ -22,10 +22,12 @@ export type MonacoEditorProps = {
     onEditorStartDone?: (editorApp?: EditorApp) => void;
     onLanguageClientsStartDone?: (lcsManager: LanguageClientManager) => void;
     onTextChanged?: (textChanges: TextContents) => void;
-    onConfigProcessed?: (editorApp?: EditorApp) => void;
+    onConfigProcessed?: (result: { textUpdated: boolean, modelUpdated: boolean}, editorApp?: EditorApp) => void;
```
martin-fleck-at: nit: I think it would be nice to have a type for the result as adopters might use this as well. 
Maybe we could even pack all three properties (including `editorApp`) into the same result type. This way API changes are a little less invasive as the signature does not really change, just the single parameter might have different properties. But this is not really necessary, just an idea.

packages/wrapper-react/src/index.tsx
```diff
@@ -22,10 +22,12 @@ export type MonacoEditorProps = {
...
     onError?: (error: Error) => void;
     onDisposeEditor?: () => void;
     onDisposeLanguageClient?: () => void;
+    toggleReprocessConfig?: boolean;
```
martin-fleck-at: nit: By now we got a lot of properties, might be nice to add some documentation to them since not all of them might be super obvious in what they do and what the effects are.

martin-fleck-at (approved):
I can confirm that issue #996 is resolved for me with that one. I could not longer reproduce the fast typing issue. Great work, thank you very much @kaisalmen!

I did add some minor comments but nothing that should prevent this PR from being merged.

kaisalmen:
@martin-fleck-at thanks for the useful comments. I will incorporate your feedback tomorrow.

packages/wrapper-react/src/index.tsx
kaisalmen: Yes, you are right. I will fix that with the next commit.

packages/wrapper-react/src/index.tsx
kaisalmen: Yes, I will add a type and integrate editorApp 👍

packages/wrapper-react/src/index.tsx
kaisalmen: Good point. Will do.

kaisalmen:
@CGNonofr and @martin-fleck-at I have implemented the review comments. Additionally, I changed `triggerReprocessConfig` to a number. Effectively, the config is only re-processed if you increment it. Also, `enforceLanguageClientDispose` does not lead to restart of the language client as soon as you type something.

New next releases are available:
https://www.npmjs.com/package/monaco-languageclient/v/10.4.0-next.2
https://www.npmjs.com/package/@typefox/monaco-editor-react/v/7.4.0-next.2

Issue #940, opened 2025-07-08
Issues with Angular 18


ChrisGilchrist:
Hi, 

I've hooked up the monaco language client in my Angular 18 application and the majority of it is working however I'm getting various 404 errors. I have used the verify/angular demo as a base to wire this up and so I tried cloning that project to run it locally and I ran into the same issues.

Any idea what is causing this and how can I resolve the url issues?

<img width="1156" height="371" alt="Image" src="https://github.com/user-attachments/assets/8169cc92-8e92-4767-b1ca-06e043ac14e1" />

kaisalmen:
Hi @ChrisGilchrist our Angular verification example is currently broken. We will try to fix it [after the the next major release](https://github.com/TypeFox/monaco-languageclient/issues/953).

Using our library stack with pure webpack or with vite/esbuild/rollup is possible, but we ran into issue we weren't able to overcome with the Angular build tooling. The investigation or problem resolution is not trivial and time consuming and has been pushed back.

faizanazim11:
Hi @kaisalmen @CGNonofr - Any update on this fix. I am working on Angular 19 and I am also facing the same issue while integrating Monaco language client and monaco-editor. If the angular example can be fixed it will be a really helpful.

kaisalmen:
Hi @faizanazim11 sorry, no news here. I need to start over with a very simple example and asses the status.

faizanazim11:
@kaisalmen - Until the issue is resolved can you provide any older version example which might be in working condition as this is required for our production features to work properly. It will be helpful until you have time to fix the current version.

kaisalmen:
@faizanazim11 it may work with custom webpack config. I am testing something right now: https://github.com/TypeFox/monaco-languageclient/tree/ng-experiments

Once rolldown-vite fixes [this issue](https://github.com/TypeFox/monaco-languageclient/issues/968#issuecomment-3484854457) we should be able to overcome the whole problem.

kaisalmen:
I have made progress here. When I use a pre-bundled version, the I get rid of most problem, but I end up having [this problem](https://github.com/TypeFox/monaco-languageclient/discussions/1000) as well.

jhk-mjolner:
Hi @kaisalmen 🙂,
What is the status of this at the moment?

I'm trying to upgrade my frontend dependencies and switch from webpack to Angular's esbuild and I'm getting this problem as well.
[...]

Issue #841, opened 2025-02-04
Error in Angular Example : Unable to load extension-file://vscode.theme-defaults/extension/themes/dark_modern.json: Not Found


rawatSingh2022:
[...]

Issue #828, opened 2025-01-20
Unable to start Angular verification example for monaco-languageclient 9.1.0


chinnychinchin:

chinnychinchin:
![Image](https://github.com/user-attachments/assets/06b62290-11a2-4891-8073-344032e96d52)

Encountered these errors during npm run build

![Image](https://github.com/user-attachments/assets/e6e36d47-ee6f-4098-b185-c27897fb935f)

kaisalmen:
Hi @chinnychinchin the builds are working, see here: https://github.com/TypeFox/monaco-languageclient/actions
Did you do any local modifications? Where there problems/warnings when you executed npm install?

kaisalmen:
Angular verification example is currently broken, see also: https://github.com/TypeFox/monaco-languageclient/issues/940#issuecomment-3264952942

PR #1001, opened 2025-12-02
Re-activate angular verification example

kaisalmen:
Now `@analogjs/vite-plugin-angular` is used instead of the Angular builder. The angular builder or both custom esbuilder or the webpack builder did not work. This builder plugin from [Analog](https://analogjs.org/) just makes use of vite and exposes all possibilities we are familiar with.

The example code has been simplified again during the process. It now again contains the JSON client that is used in other examples as well.

Resolves #1000, #940, #841, #828

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ dist
 lib
 node_modules
 *.tsbuildinfo
-.angular
 __screenshots__
 .chrome
 
--- a/verify/angular/.gitignore
+++ b/verify/angular/.gitignore
@@ -1,7 +1,7 @@
+.angular
 bundle
 dist
 lib
 node_modules
 *.tsbuildinfo
 production
-package-lock.json


verify/angular/esbuild.config.mjs
```diff
@@ -0,0 +1,3 @@
+import importMetaUrlPlugin from '@codingame/esbuild-import-meta-url-plugin';
+
+export default importMetaUrlPlugin;
```
CGNonofr: not familiar with angular but... what is the point of this file? :thinking:

verify/ng/.gitignore
```diff
@@ -0,0 +1 @@
+package-lock.json
```
CGNonofr: is it on purpose?

CGNonofr (approved):
LGTM for what I'm able to understand

verify/angular/esbuild.config.mjs
kaisalmen: It was an experiment I disregarded. You can provide esbuild plugins for the custom esbuild builder plugin. The plugin was loaded by the builder, but not applied.

verify/ng/.gitignore
kaisalmen: I tried the old webpack conig and used this as the base for the changes. There the lock file was ignored. I overlooked this. Thanks. Will be removed in upcoming commit.

PR #1006, opened 2025-12-23
LanguageClient config enhancements, plus dynamic loading of extension-service-override

### LINKED ISSUE #1004: Language Client features register
Hello, I need to access to the Language Client to set some `experimental` flags on request before it sent to the Language Server to initiate the connection. The current way is using a class that extended from `StaticFeature` to set these value.

But with `MonacoEditorReactComp` there is no way to access the client before `LanguageClientWrapper` call the `start()` method.

Image

I need a way to call this `registerFeature`

```js
client.registerFeature(new MyFeature())
````

kaisalmen:
This also updates dependencies and prepares the next release.

This change also adds configuration enhancement and code changes that allow to load the extension-service-override to load later. It is only performed dynamically in any case (worker extension host or not).

Fixes #1004


kaisalmen:
@CGNonofr I will merge this myself. Happy holidays!🎄🙂

PR #1011, opened 2026-01-13
Remove own ConsoleLogger implementation in mlc/common

### LINKED ISSUE #1010: Localization not applying to all strings
I tried running the Langium example -> "Localized Versions" (updated main branch). Only a few strings are updating. I would have expected all the strings to get localized.

Image

-----

- This menu is also not being localized
Image

Getting console error:
>Some parts of VSCode are already initialized, make sure the language pack is loaded before anything else or some translations will be missing

kaisalmen:
It lead to too early api init. Aligned test. The apiWrapper ensures that logging is turned off per default.

Closes #1010


CGNonofr (approved):
LGTM

PR #1035, opened 2026-03-18
Update dependencies once more and update GitHub action versions

kaisalmen:
The mise action was still relying on an old node version.


kaisalmen:
@CGNonofr will there be a v24 soon? We can just leave that open until it is there.

CGNonofr:
> @CGNonofr will there be a v24 soon? We can just leave that open until it is there.

You mean a v29 with vscode 1.112? probably not, it only contains minor changes

kaisalmen:
> You mean a v29 with vscode 1.112? probably not, it only contains minor changes

Yes, sorry. I meant 29/1.112. Then this can be merged from my point of view

CGNonofr (approved):

PR #1040, opened 2026-06-17
Make examples package private, update verification lock files

kaisalmen:
Use only use examples package relatively. There will be no more releases from now on.

Re-generate all lock files in verification examples to overcome dependency security warnings (many outdated packages).

### DIFF
--- a/verify/webpack/webpack.config.js
+++ b/verify/webpack/webpack.config.js
@@ -38,15 +38,11 @@ const config = {
   },
   target: 'web',
   resolve: {
-    extensions: ['.ts', '.js', '.json', '.ttf'],
-    fallback: {
-      'fs/promises': false
-    }
+    extensions: ['.ts', '.js', '.json', '.ttf']
   },
   plugins: [
-    // intercept the "node:" URI scheme and strip it
-    new webpack.NormalModuleReplacementPlugin(/^node:/, (resource) => {
-      resource.request = resource.request.replace(/^node:/, '');
+    new webpack.IgnorePlugin({
+      resourceRegExp: /^node:fs/
     })
   ],
   mode: 'development',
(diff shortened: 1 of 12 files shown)


verify/webpack/webpack.config.js
```diff
@@ -39,11 +40,15 @@ const config = {
...
-      module: false,
-      vm: false
+      'fs/promises': false
     }
   },
+  plugins: [
+    // intercept the "node:" URI scheme and strip it
+    new webpack.NormalModuleReplacementPlugin(/^node:/, (resource) => {
+      resource.request = resource.request.replace(/^node:/, '');
+    })
```
CGNonofr, line 46: Why is it needed? importing node module seems a bad idea anyway 🤔

kaisalmen: Webpack picks up a `node:fs/promises` import from `@vscode/diff/dist/wire.js` . It is this function:
```js
async function initWasm() {
    if (wasmModule) {
        return wasmModule;
    }
    if (initPromise) {
        return initPromise;
    }
    initPromise = (async () => {
        const wasmUrl = new URL(new URL('diff_wasm_bg-c6YkaZRL.wasm', import.meta.url).href);
        // Node's undici fetch does not support file:// URLs, so read bytes ourselves.
        let source = wasmUrl;
        if (typeof process !== 'undefined' && process.versions?.node) {
            const { readFile } = await import('node:fs/promises');
            source = await readFile(wasmUrl);
        }
        await __wbg_init({ module_or_path: source });
        init();
        wasmModule = wasm$1;
        return wasmModule;
    })();
    return initPromise;
}
```
Interestingly vite automatically detects it and externalizes it.

kaisalmen: That dependency / code must have changed in the last month since I tried the webpack example.

CGNonofr: Yes, @vscode/diff is a brand new package of the 1.124

What I did is using the IgnorePlugin instead:
```typescript
    new webpack.IgnorePlugin({
      resourceRegExp: /^node:fs/,
    })
```

kaisalmen: Thanks, I will adapt. Commit is coming soon.

CGNonofr (approved):

PR #1041, opened 2026-06-17
Use locked installation with verify packages.

kaisalmen:
Use further overrides to remove further dependabot alerts.

Hopefully the last PR related to dependencies and security for today. I want to reduce the dependabot alert list and activate it again afterwards

### DIFF
--- a/verify/angular/package.json
+++ b/verify/angular/package.json
@@ -4,8 +4,8 @@
   "private": "true",
   "type": "module",
   "scripts": {
-    "verify": "npm ci --immutable && npm run start",
-    "verify:ci": "npm install --immutable && npm run build",
+    "verify": "npm ci && npm run start",
+    "verify:ci": "npm ci && npm run build",
     "clean": "shx rm -fr dist *.tsbuildinfo src/assets",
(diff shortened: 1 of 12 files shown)


verify/angular/package.json
```diff
@@ -4,8 +4,8 @@
   "private": "true",
   "type": "module",
   "scripts": {
-    "verify": "npm install && npm run start",
-    "verify:ci": "npm install && npm run build",
+    "verify": "npm ci --immutable && npm run start",
```
CGNonofr: I don't find any doc on the `--immutable` flag 🤔

CGNonofr: Oh it's a yarn-only flag

kaisalmen: Yes, we use yarn 4 here.

CGNonofr: You use yarn via npm? :thinking:  I'm confused

kaisalmen: No, sorry, that was  a mistake. Well-spotted. Corrected :arrow_down:

CGNonofr (approved):

PR #1044, opened 2026-06-22
Migrate over Langium Generation and Integration Guides

### LINKED ISSUE #868: Integrate Langium project generation and documentation
We integrated this into the Langium generator, but it was usually outdated.
The idea is to add generation instructions (code and doc) here and keep it in sync (see [issue](https://github.com/eclipse-langium/langium-website/issues/274)). This is less effort for all.

montymxb:
Closes #868 by bringing over a pair of Langium guides that demonstrated how to integrate & leverage Langium + Monaco. Those have been out of date on the Langium side, and needed a bit of tlc, so I refreshed them and brought them over here. Most of the content is from the original tutorial, but rewritten and adapted for the new MLC api and any changes since Langium 3.X.

The updated sections have been setup so they should follow after the practical examples section, by building up into a Langium context. The first guide is about setting up and integrating an LS worker built on the Langium side (very similar to the existing guide). The second one has been adapted to instead focus on how to work with notifications & requests instead of 'generation' by itself. This broadens the scope of application a bit, so it should be clearer that although generation is a use case, it's not the only thing you could be doing there.

Lastly this keeps with the existing trend of using the [langium-minilogo](https://www.npmjs.com/package/langium-minilogo) example language. However the guides have both been updated to walk through the process generally as well as concretely via minilogo, so it should be helpful for users who want to know the general flow or see a concrete example.

There's a new minilogo example added in as well for reference, which consumes a pre-built LS rather than integrate the source directly. I updated the language up to Langium v4.3.0 as well, after it had been sitting for a couple years collecting a bit of dust 🧹 .


packages/examples/ghp_minilogo.html
```diff
@@ -0,0 +1,17 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>MiniLogo Client & Language Server (Worker)</title>
+    <link rel="stylesheet" href="style.css" />
+  </head>
+
+  <body>
+    <div class="exampleHeadelineDiv">
+      <b class="exampleHeadeline">MiniLogo Client & Language Server (Worker)</b> - [<a href="../../index.html">Back to Index</a>]
```
kaisalmen: This link must be:
```suggestion
      <b class="exampleHeadeline">MiniLogo Client & Language Server (Worker)</b> - [<a href="./index.html">Back to Index</a>]
```

docs/guides/langium/running-langium-ls-in-browser.md
```diff
@@ -0,0 +1,473 @@
...
+import { startLanguageServer } from 'langium/lsp';
+import {
+  BrowserMessageReader,
+  BrowserMessageWriter,
+  createConnection
+} from 'vscode-languageserver/browser';
+// your services import will differ based on your language
+import { createMyLanguageServices } from './my-language-module.js';
+
+declare const self: DedicatedWorkerGlobalScope;
```
kaisalmen, line 42: Btw, with this option, you do not need to use `WebWorker` in the compiler options lib:
```ts
/// <reference lib="WebWorker" />
```

kaisalmen (changes requested):
Thank you very much. 🎉 
One link needs to be corrected. Then it is ready to go.
The other comment is just for info.

kaisalmen (approved):
LGTM

PR #1051, opened 2026-07-09
Introduce Worker Label type definition for extended and classic

### LINKED ISSUE #1038: workerLoaders should use typed worker labels instead of string
We recently upgraded `monaco-languageclient` from `10.1.0` to `10.7.0`. In our configuration, we were registering workers as follows:

```ts
useWorkerFactory({
  workerLoaders: {
    TextEditorWorker: () => params.workers.textEditorWorker,
    TextMateWorker: () => params.workers.textMateWorker,
  },
  logger,
});
```

In `10.7.0`, `TextEditorWorker` appears to have been replaced by `editorWorkerService` in the default worker loader mapping [here](https://github.com/TypeFox/monaco-languageclient/blob/ca8a73c1d6040adbb5e894f86795697b27252dce/packages/client/src/worker/index.ts#L50-L52). After the upgrade, this mismatch did not produce any type error, so it took us a while to identify that the editor worker configuration was no longer being matched and the corresponding work was falling back to the UI thread.

At the moment, `workerLoaders` is typed as `Partial<Record<string, WorkerLoader>>`. Would it be possible to narrow this type so that only supported worker names are allowed? That would help catch renamed or removed worker keys such as this one at compile time instead of surfacing later as a runtime behavior change.

kaisalmen:
Upside: Clearer what are allowed worker names.
Downside: Needs to be maintained, but I think it's ok.

Fixes #1038


kaisalmen:
@CGNonofr are you ok with these changes?

packages/client/src/worker/index.ts
```diff
@@ -18,26 +18,39 @@ export class Worker {
...
 export type WorkerLoader = (() => Worker) | undefined;
 
+export type PossibleWorkerLabelsExtended =
+  | 'editorWorkerService'
+  | 'extensionHostWorkerMain'
+  | 'TextMateWorker'
+  | 'OutputLinkDetectionWorker'
+  | 'LanguageDetectionWorker'
+  | 'NotebookEditorWorker'
+  | 'LocalFileSearchWorker';
```
CGNonofr, line 30: I think some libraries relies on the monaco worker loading mechanism, like monaco-graphql, so this list should probably remain extensible 🤔

CGNonofr (commented):
What about doing that on the monaco-vscode-api side? :thinking:

packages/client/src/worker/index.ts
kaisalmen: Yes, but allowing string as third type breaks the constraint, doesn't it? Or will this work? I need to test it.

kaisalmen: Ok, that will work. I will come up with a proposal and yes, maybe this can be moved to monaco-vscode-api.

kaisalmen: Done :arrow_down:

kaisalmen:
> What about doing that on the monaco-vscode-api side? 🤔

Should we move the type definitions to monaco-vscode-api?

CGNonofr:
> > What about doing that on the monaco-vscode-api side? 🤔
> 
> Should we move the type definitions to monaco-vscode-api?

It probably makes more sense :thinking:

kaisalmen:
@CGNonofr what about we merge this first and then open a new PR in mva. And drop the definitions again with the new release. That should be straight-forward.

CGNonofr (approved):

CGNonofr:
> @CGNonofr what about we merge this first and then open a new PR in mva. And drop the definitions again with the new release. That should be straight-forward.

It looks more complicated to me but feel free to do it as you wish :)

PR #1055, opened 2026-07-22
Update monaco-vscode-update to v36

kaisalmen:
The two worker tests are in its own test tag. `fileParallelism` is only used there in addition.
It would fix the problem if globally applied with using tags, but speed is much worse.


vite.config.ts
```diff
@@ -37,6 +37,7 @@ export const definedViteConfig = defineConfig({
   },
   server: {
     port: 20001,
+    host: '0.0.0.0',
```
CGNonofr, line 40: It would be a bit simpler to review if unrelated changes would be in separate commits :)

CGNonofr (approved):

vite.config.ts
kaisalmen: Yes, I try to do that. That was a left-over. I am running all my dev envs in containers now. That's why a added that, btw.

PR #1061, opened 2026-09-10
Rework connection usage in LanguageClientWrapper

kaisalmen:
The current connection handling is a mess.

This PR introduces a base interface, a support class (`LanguageClientConnectionSupport` and `LanguageClientConnectionRealization`) and contains implementations (= the websocket and worker handling moved to their own classes.

This had some impact on the api and therefore the examples.

This PR is a first step. I am already working in a socket.io based connection realization and commendable server for language server tests. This is required to test and verify re-connection and timeouts which is currently not really working. I want to introduce and fix things before final version 11. 

All these idea were there for longer, but I didn't really got to it.


packages/client/src/wrapper/lcwrapper.ts
```diff
@@ -39,32 +42,48 @@ export class LanguageClientWrapper {
...
+   * Optionally, you can init independently before start.
+   * This allows for example to get access to MessageTransports or the Worker before starting the language client.
+   */
+  async init(): Promise<void> {
+    this.connectionRealization.connected = () => {
+      this.connectionEstablished.resolve(true);
+    };
+    this.connectionRealization.disconnected = async () => {
+      await this.dispose();
+    };
```
CGNonofr, line 62: what about using an event emitter instead? 🤔

packages/vscode-ws-jsonrpc/src/socket/reader.ts
```diff
@@ -13,9 +14,9 @@ export class WebSocketMessageReader extends AbstractMessageReader implements Mes
   // oxlint-disable-next-line @typescript-eslint/no-explicit-any
   protected readonly events: Array<{ message?: any; error?: any }> = [];
 
-  constructor(socket: IWebSocket) {
+  constructor(webSocket: WebSocket | IWebSocket) {
     super();
-    this.socket = socket;
+    this.socket = Object.hasOwn(webSocket, '$type') ? (webSocket as IWebSocket) : toSocket(webSocket as WebSocket);
```
CGNonofr: nitpicking: instead of casting, you can use a typeguard function

vite.config.ts
```diff
@@ -82,6 +82,21 @@ export const definedViteConfig = defineConfig({
...
+          if (req.url?.includes('packages/client/test/support/unreachableFile.ts') === true) {
+            res.statusCode = 404;
+            res.end('Not Found');
+            return;
+          }
+
+          next();
+        });
+      }
+    },
```
CGNonofr, line 99: what is this for?

CGNonofr (approved):
A lot of changes, but LGTM

> I am already working in a socket.io based connection realization and commendable server for language server tests.

You may want to use engine.io instead maybe

vite.config.ts
kaisalmen: The vite dev server returns index.html when a URL is not reachable. That prevents that for the specific file. I need that for a Worker test where the worker URL is invalid.
Do you have another idea how to do this?

CGNonofr: Ok, it's fine

packages/vscode-ws-jsonrpc/src/socket/reader.ts
kaisalmen: Will do.

packages/client/src/wrapper/lcwrapper.ts
kaisalmen: Does not make sense. I removed the connect callback, though. The start works with a promise now and I hink error handling is clearer now.

kaisalmen:
> You may want to use engine.io instead maybe

@CGNonofr you mean the foundation of socket.io, right? Yes, maybe that could be an option.

